PR #49, opened 2021-08-20
Develop

SashaRudenko:


SashaLifashkin:
Классное решение!

src/groupByKey.ts
```diff
@@ -1,7 +1,20 @@
...
+  return items.reduce(
+    (accumulator, item) => {
+      const value = item[field];
+
+      accumulator[value] = (accumulator[value] || []).concat(item);
+
+      return accumulator;
+    },
+    {} as Record<T[K], T[]>
+  );
```
VinnieJ-2k20, line 19: Don't overuse generic types. Use them only when necessary. 
Here it would be much better just to introduce additional type, like this:
```
type ItemsGroup<T> = {
  [key: string]: T[]
};
```
And then use it:
```suggestion
export function groupByKey(
  items: Item[],
  field: keyof Item,
): ItemGroup<Item>{
  return items.reduce(
    (accumulator, item) => {
      const value = item[field];

      accumulator[value] = (accumulator[value] || []).concat(item);

      return accumulator;
    },
    {},
  );
```

VinnieJ-2k20 (changes requested):
Almost done, let's simplify your code.

VinnieJ-2k20 (approved):
Good job 👍

PR #56, opened 2021-08-23
solution

Vova-Stoliar:


nastyyvolk (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ interface Item {
   [key: string]: any;
 }
 
-export function groupByKey(items, field) {
+export function groupByKey(items: Item[], field: string): Item {
   // write code here;
+  const result: Item = {};
+
+  items.forEach((info) => {
+    if (result[info[field]] === undefined) {
```
VinnieJ-2k20, line 10: We know that value is undefined or array, so we can use shorter condition
```suggestion
    if (!result[info[field]]) {
```

VinnieJ-2k20 (approved):
Good job 👍

PR #57, opened 2021-08-23
Group By Key solution

TarasHoliuk:


sasha-morozov (approved):
As for me, the code is good and there is nothing redundant!)

Denchepornyuk (approved):

PR #69, opened 2021-08-27
solution

evgen1sider:


src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ interface Item {
...
-  // write code here;
+interface FilterBy {
+  [key: string]: Item[];
+}
+
+export function groupByKey(items: Item[], field: keyof Item): FilterBy {
+  const filteredByKey: FilterBy = {};
+
+  for (let i: number = 0; i < items.length; i += 1) {
+    if (items[i][field] in filteredByKey) {
```
varseniuk, line 13:
```suggestion
  items.forEach((item: Item) {
    if (item[field] in filteredByKey) {
```

You could use `forEach` instead of `for` loop and make your code more readable

varseniuk (approved):
Good job, but pay attention to my comment

PR #72, opened 2021-08-27
Done

karolina-nad:


AndriyNikorych (commented):
Good job👍

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ interface Item {
   [key: string]: any;
 }
 
-export function groupByKey(items, field) {
-  // write code here;
+export function groupByKey(items: Item[], field: keyof Item):Item {
+  const resultList: Item = {};
```
varseniuk, line 6: The result should be not of an `Item` type. It should be an array of 'items'

Create the interface 
```
{
  [key: string] : Item[],
}
```

and use it as a result type

varseniuk (changes requested):
All is good, but let's use proper types

src/groupByKey.ts
```diff
@@ -1,7 +1,21 @@
 interface Item {
-  [key: string]: any;
+  [key: string] : any,
 }
 
-export function groupByKey(items, field) {
-  // write code here;
+interface GropedItems {
```
varseniuk, line 5:
```suggestion
interface GroupedItems {
```
typo

src/groupByKey.ts
```diff
@@ -1,7 +1,21 @@
...
+  [key: string] : Item[],
+}
+
+export function groupByKey(items: Item[], field: keyof Item):Item {
```
varseniuk, line 9:
```suggestion
export function groupByKey(items: Item[], field: keyof Item): GroupedItems {
```
The function should return the object of proper type

varseniuk (changes requested):
One small fix left

varseniuk (approved):
🔥

PR #78, opened 2021-08-29
Solution Group By Key

AndriyNikorych:


karolina-nad:
Good job!

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ interface Item {
   [key: string]: any;
 }
 
-export function groupByKey(items, field) {
-  // write code here;
+export function groupByKey(items: Item[], field: keyof Item): Item {
+  const sortByKey: Item = {};
```
varseniuk, line 6: The result should be not of an Item type. Each key should be an array of 'items' (not 'any')

Create the interface
```
{
  [key: string] : Item[],
}
```
and use it as a result type

varseniuk (changes requested):
All is good but let's use proper types

src/groupByKey.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+exports.__esModule = true;
```
varseniuk, line 2: why do you need this file? 🤔

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ interface Item {
...
 }
 
-export function groupByKey(items, field) {
-  // write code here;
+interface SortItem {
+  [key: string]: Item[];
+}
+
+export function groupByKey(items: Item[], field: keyof Item): SortItem {
+  const sortByKey: Item = { };
```
varseniuk, line 10:
```suggestion
  const sortByKey: SortItem = { };
```
Your function returns the type `SortItem`, but your `sortByKey` variable is of another type. Fix this

varseniuk, line 10:
```suggestion
  const sortByKey: Item = {};
```

varseniuk (changes requested):

varseniuk (approved):

PR #79, opened 2021-08-29
TypeScript - Extras - Group by key

george-w-boo:


UlaDutchak (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ interface Item {
   [key: string]: any;
 }
 
-export function groupByKey(items, field) {
+export function groupByKey(items: Item[], field: keyof Item): Item {
   // write code here;
+  const result: Item = {};
```
varseniuk, line 7: The result should be not of an 'Item' type. Each key should be an array of 'items' (not 'any')

Create the interface
```
{
  [key: string] : Item[],
}
```
and use it as a result type

varseniuk (changes requested):
All is good but let's use proper result type

src/groupByKey.ts
george-w-boo: I've just fixed it. Thank you :)

src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ interface Item {
...
 
-export function groupByKey(items, field) {
+interface Result {
+  [key: string]: Item[];
+}
+
+export function groupByKey(items: Item[], field: keyof Item): Result {
   // write code here;
+
+  const result: Item = {};
```
varseniuk, line 12:
```suggestion
  const result: Result = {};
```
This variable should be of a proper type as well

varseniuk (changes requested):

src/groupByKey.ts
george-w-boo: I couldn't agree more, thank you ;)

varseniuk (approved):
🔥

PR #80, opened 2021-08-29
first try

zadorojniiarsen:


src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ interface Item {
   [key: string]: any;
 }
 
-export function groupByKey(items, field) {
-  // write code here;
+export function groupByKey(items: Item[], field: keyof Item) :Item {
+  const result: Item = {};
```
varseniuk, line 6:
```suggestion
  const result: Item = {};
```
The result should be not of an 'Item' type. Each key should be an array of 'items' (not 'any')

Create the interface
```
{
  [key: string] : Item[],
}
```
and use it as a result type
Also, `result` is a bad naming

varseniuk (changes requested):

varseniuk (approved):
🔥

PR #81, opened 2021-08-29
Develop

Oleksandr-Zhitenko:


zadorojniiarsen (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ interface Item {
   [key: string]: any;
 }
 
-export function groupByKey(items, field) {
+export function groupByKey(items: Item[], field: string): object {
   // write code here;
+  const obj: Item = {};
```
varseniuk, line 7: The result should be not of an `Item` type. Each key should be an array of 'items' (not 'any')

Create the interface
```
{
  [key: string] : Item[],
}
```
and use it as a result type.

Also, `obj` is a very bad name

varseniuk (changes requested):
All is good but let's use proper types and namings

varseniuk (approved):
👍

PR #116, opened 2021-10-06
Solution

YulyaVaV:


src/groupByKey.ts
```diff
@@ -1,7 +1,20 @@
...
+  [key:string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: Item[], key: keyof Item):GroupsMap<Item> {
+  const groups: GroupsMap<Item> = {};
+
+  items.forEach((obj) => {
+    if (!Object.prototype.hasOwnProperty.call(groups, obj[key])) {
```
aliceblunt8, line 13: let's just check if this key is in object (use in operator or object[key]), I think Object.prototype.hasOwnProperty.call looks strange)

aliceblunt8 (approved):

PR #161, opened 2021-11-24
solution was added

JamesSunderlanddd:


src/groupByKey.ts
```diff
@@ -1,7 +1,16 @@
-type GroupsMap<T> = {
-  [key: string]: T[];
-};
+interface Item {
+  [key: string]: any;
```
dmtrhrytsak, line 2: By using `any` you miss the point of typescript) It should be of generic type

`  [key: string]: T[];`

dmtrhrytsak (approved):
Keys are grouped, so I'm satisfied🤩

src/groupByKey.ts
JamesSunderlanddd: @dmtrhrytsak  I was just so fed up with the complexity of some typescript stuff that I learned recently, that I wanted to simplify things a bit) But you are right, I better change it)

dmtrhrytsak: No, that's okay, I understand you) You deserve resting not this whole typescript thing! :D

src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof {}): GroupsMap<T> {
```
illiarerun, line 5: Keys of plain object? Doesn't look right
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
```

illiarerun (changes requested):
Good job 🔥  Let's fix one little issue:

illiarerun (approved):
Great job 🔥

PR #164, opened 2021-11-29
Solution

Val-space:


dmtrhrytsak (approved):
It’s simple, straightforward, elegant, and solves all the problems🎀

aionova (approved):
Nice!

PR #168, opened 2021-11-30
Group By Key

Vitaliy-0:


msaikun (approved):
Good job!

src/groupByKey.ts
```diff
@@ -1,7 +1,26 @@
...
 type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> | undefined {
```
ihor-jpeg, line 9: when does it return undefined?

src/groupByKey.ts
```diff
@@ -1,7 +1,26 @@
+/* eslint-disable no-prototype-builtins */
 type GroupsMap<T> = {
   [key: string]: T[];
```
ihor-jpeg, line 3: to make it even better, you can make it so that the key can only be one of the keys of T.
```suggestion
  [key in keyof T]?: T[];
```

ihor-jpeg: your function already takes care of that, just to double-check it here

ihor-jpeg (approved):
Good work. Have a look at the comments

PR #171, opened 2021-11-30
new solution

mariia-bila:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<Item>(items:Item[], key:keyof Item):GroupsMap<Item> {
+  const obj:GroupsMap<Item> = {};
+
+  items.forEach((item) => {
+    const keyValue = String(item[key]);
+
+    if (!obj[keyValue]) {
+      obj[String(item[key])] = items.filter((el) => el[key] === item[key]);
+    }
```
dmtrhrytsak, line 13: How about rewriting it this way?)

```suggestion
    if (!obj[keyValue]) {
      obj[keyValue] = [item];
    } else {
      obj[keyValue].push(item);
    }
```

dmtrhrytsak (approved):
Brilliant solution! You kill it every time🎯

src/groupByKey.ts
aionova: agree with Dima. This line is quite complicated to read

aionova (approved):

PR #172, opened 2021-12-01
solution: ts_group_by_key task

lanebx:


ImIkator (approved):
Good solution !

src/groupByKey.ts
```diff
@@ -2,6 +2,24 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce(
+    (acum: {}, item: T) => {
+      const NameKey = String(item[key]);
```
aionova, line 8: we make a first letter capital only if it's a type. We don't use them for types
```suggestion
      const nameKey = String(item[key]);
```

src/groupByKey.ts
```diff
@@ -2,6 +2,24 @@ type GroupsMap<T> = {
...
+
+      if (NameKey in acum) {
+        return acum;
+      }
+
+      const newT = {
```
aionova, line 14: what does newT mean? this name of variable doesn't tell me anything about what's inside of it

aionova (changes requested):

aionova (approved):
Nice!

PR #183, opened 2021-12-01
Solution

IrynaHerasymchuk:


SerhiiKirik (approved):
Good job 🔥

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+  items.forEach((item) => {
+    const resultKey = String(item[key]);
+
+    if (grouppedObjects[resultKey]) {
+      grouppedObjects[resultKey].push(item);
+    } else {
+      grouppedObjects[resultKey] = [item];
+    }
+
+    return grouppedObjects;
```
aionova, line 17: it looks like you wanted to use reduce, but used forEach instead, did you? Why did you need the second parameter for forEach?

aionova (changes requested):
you use forEach in a weird way)

src/groupByKey.ts
IrynaHerasymchuk: My bad, forgot to remove🤦

aionova: there is also no need to return something from the callback of foreach

aionova (approved):

PR #206, opened 2021-12-30
works

KHTRE:


pasha-melnyk (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item) => {
+    if (!keys.includes(item[key])) {
+      keys.push(item[key]);
+    }
+  });
+
+  keys.forEach((oneKey) => {
+    result[oneKey] = items.filter((a) => a[key] === oneKey);
+  });
```
Elena-brv, line 17: you can do this task with 1 cycle, not you are doing 2 forEach + 1 filter

Elena-brv (changes requested):

danheim (approved):

PR #208, opened 2022-01-06
solution

samson-sim:


Basil-Panasiuk (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
...
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: object[], key: string): GroupsMap<object> {
+  const groupItems: GroupsMap<object> = {};
+
+  items.forEach((item: object) => {
+    groupItems[item[key]] = items.filter(
+      (obj): boolean => obj[key] === item[key],
+    );
+  });
```
Olexandr6, line 12: Don't use the `object` type, use generic types instead!

Olexandr6 (changes requested):
You should use generic types in this task

Olexandr6 (approved):
Good job 👍

PR #211, opened 2022-01-11
My solution

SavelovSerhii:


Basil-Panasiuk (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: GroupsMap<T>[], key: string): object {
+  const result: object = {};
```
Olexandr6, line 6: Don't use `object` type. Use generic types instead

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: GroupsMap<T>[], key: string): object {
```
Olexandr6, line 5: `items` will be `T` type and you should return an object of type `GroupsMap<T>`

Olexandr6 (changes requested):
You should use generic types in this task

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+  let current;
+
+  for (let i = 0; i < items.length; i += 1) {
+    current = String(items[i][key]);
```
Olexandr6, line 10:
```suggestion
    groupKey = String(items[i][key]);
```

Olexandr6 (approved):

PR #215, opened 2022-01-11
solution

Basil-Panasiuk:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: object[], key: string): GroupsMap<object> {
+  const groupedData: GroupsMap<object> = {};
+
```
andrii-koptiev, line 7: Not use generic here. you can try  groupByKey<Item> (arguments)....

andrii-koptiev, line 7: You not use generic here.  to use you should do it like this http://joxi.ru/4AkdqYQT0Zb3Dr

destroythemoon (approved):

PR #219, opened 2022-01-12
group by key

Toros-Zz:


bcmp4p1 (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const group: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const groupKey = `${item[key]}`;
+
+    if (group[groupKey] === undefined) {
```
Olexandr6, line 11: if the value is equal to `undefined`, it will be a runtime error.
Use another way to check, if a property exists in an object

Olexandr6 (changes requested):
Well done 👍 But some thing should be fixed

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const group: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const groupKey = `${item[key]}`;
+
+    if (!group[groupKey]) {
```
Olexandr6, line 11: You've done the same condition, but in the shortened form) You should use `hadOwnPropery()` in such case. But okay, we'll omit this)

Olexandr6 (approved):

PR #233, opened 2022-02-08
empty... for now...

tvoyammn:


src/groupByKey.ts
```diff
@@ -2,6 +2,12 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: object[], key: string): GroupsMap<object> {
```
trtskvalerie, line 5: я думала так....
```suggestion
export function groupByKey(items: object[], key: keyof T): GroupsMap<object> {
```

trtskvalerie (approved):
damnnnnnnnnnnnnnnnnnnnn

src/groupByKey.ts
Anna-Tkachenko: Don't use object and string types
Create some type instead of object and keyof YourType for key argument

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
🔥👌 🚀

PR #237, opened 2022-02-09
solution

mykhailoyanush:


013g (approved):
Well done!

src/groupByKey.ts
```diff
@@ -1,7 +1,22 @@
...
+
+export function groupByKey(items: Item[], key: keyof Item): GroupsMap {
+  const groups: GroupsMap = {};
+
+  items.forEach((item) => {
+    if (groups[item[key]]) {
+      groups[item[key]].push(item);
+    } else {
+      groups[item[key]] = [];
+      groups[item[key]].push(item);
```
Anna-Tkachenko, line 17: Don't repeat yourself: you can call push method only once

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #252, opened 2022-02-21
Solution

Anna-Morozevych:


src/groupByKey.ts
```diff
@@ -2,6 +2,12 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    result[String(item[key])] = items.filter((obj) => obj[key] === item[key]);
+  });
```
aliceblunt8, line 10: let's optimize your solution, now on each iteration you call filter, so your approach has quadratic complexity, let's use push instead, so the complexity will be linear

https://medium.com/@ariel.salem1989/an-easy-to-use-guide-to-big-o-time-complexity-5dcf4be8a444#:~:text=Quadratic%20Time%20Complexity%20represents%20an,iterations%20within%20the%20data%20sets.

aliceblunt8 (changes requested):

ArtemTopolskyi (approved):

PR #281, opened 2022-04-21
GroupByKey function implemented

VETALEK:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
```
ivan-kadykalo, line 5:
```suggestion
export function groupByKey<T>(items: T[], key: key of T): GroupsMap<T> {
```
It would be better to check key from your T type

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+  const groups: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    if (groups[item[key]]) {
+      groups[item[key]].push(item);
+    } else {
+      groups[item[key]] = [item];
+    }
```
ivan-kadykalo, line 13:
```suggestion
    if (!groups[item[key]]) {
      groups[item[key]] = [];
    }
    groups[item[key]].push(item);
```

And here you can avoid else statement if you will check your key on false for the first.
There can be problem with a type of `item[key]`, so you should convert it to string.

ivan-kadykalo (changes requested):

AlinaMatuschak (approved):
Nice, but also you can use reduce for it )

PR #331, opened 2022-05-24
solution

efdalyalcin:


src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(
+  items: T[],
+  key: keyof T,
+): object {
+  const grouppedObject: GroupsMap<T> = { };
```
59MrRobot, line 9: Spelling error here

meforoma (approved):
Swazi is right, but it's not a blocking tweak.
Fix in spare time @efdalyalcin

PR #333, opened 2022-05-25
Solution

amitpatiljc:


59MrRobot (approved):
Good job 👍

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const Objects: GroupsMap<T> = {};
```
meforoma, line 6: this one is quite important from convention perspective
when you capitalize variables they must be `type / interface / react component`
if it's a regular variable - first char must be lower-case
```suggestion
  const objects: GroupsMap<T> = {};
```

meforoma: and ofc apply throughout the code
remember, rename using `F2` to apply new name to all occurences

meforoma (changes requested):

src/groupByKey.ts
amitpatiljc: Done

meforoma (approved):

PR #339, opened 2022-05-27
Solution

dima-cherednyk:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export
+function groupByKey(items: object[], key: keyof object): object {
```
vkryvytskyy, line 6:
```suggestion
function groupByKey<T>(items: object[], key: keyof object): object {
```
function should have a generic type, don't use object

vkryvytskyy (changes requested):

ArtemTopolskyi (approved):
choose one variant and remove comments

PR #343, opened 2022-06-06
solution

nastia-sydorchuk:


src/groupByKey.ts
```diff
@@ -2,6 +2,23 @@ type GroupsMap<T> = {
...
+
+    if (!groupedArrays[property]) {
+      groupedArrays[property] = [];
+      groupedArrays[property].push(items[i]);
+
+      for (let c = i + 1; c < items.length; c += 1) {
+        if (String(items[c][key]) === property) {
+          groupedArrays[property].push(items[c]);
+        }
+      }
```
vkryvytskyy, line 19: You can solve this task without using nested loop :)

vkryvytskyy (changes requested):

PR #353, opened 2022-06-28
solved

mflkvsk:


GavrilyukYura (approved):
Good job

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: object[], key: keyof object):
```
Olexandr6, line 5: `object[]`, `keyof object` === any. You should avoid `object` type, case it means `any`. If you write `keyof object` it means any props but we don't need any props.
You should read more about generic types in functions - `keyof object` - https://mate-academy.github.io/fe-program/js/extra/typescript#use-type-as-a-function-parameter
```suggestion
export function groupByKey(items: object[], key: keyof object):
```

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+GroupsMap<object> {
+  const groupedObj: GroupsMap<object> = {};
```
Olexandr6, line 7: Here should be generic type not object

Olexandr6 (changes requested):

Olexandr6 (approved):

PR #355, opened 2022-06-29
my solution

Pavlo000:


src/groupByKey.ts
```diff
@@ -1,7 +1,16 @@
-type GroupsMap<T> = {
-  [key: string]: T[];
+type GroupsMap = {
+  [key: string]: [];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: GroupsMap[], key: keyof GroupsMap): object {
```
Olexandr6, line 5: You should read more about generic types in functions - https://mate-academy.github.io/fe-program/js/extra/typescript#use-type-as-a-function-parameter

G-YaSha (changes requested):
In this task you should avoid using `object` type

G-YaSha (approved):

PR #380, opened 2022-07-20
grouped by key

silchyn:


solaryasha (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    if (result[String(item[key])]) {
```
solaryasha, line 9: you are using this expression  three times - `[String(item[key])`.
Let's move it to the variable

solaryasha (approved):

PR #391, opened 2022-07-26
Solution

Kras4iy:


TwoAndMore (commented):
GJ 🔥

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const itemKey = String(item[key]);
+
+    if (!Object.prototype.hasOwnProperty.call(grouped, itemKey)) {
```
ivan-kadykalo, line 11: I think it would simpler but your solution also good👍
```suggestion
    if (!grouped[itemKey]) {
```

ivan-kadykalo (approved):

PR #396, opened 2022-07-27
Solution

LeonidPustovit:


OlgaStakhova (approved):
Як на мене гарна робота

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+
+  for (let i = 0; i < items.length; i += 1) {
+    if (!values.includes(items[i][key])) {
+      values.push(items[i][key]);
+    }
+  }
+
+  for (let i = 0; i < values.length; i += 1) {
+    obj[values[i]] = [];
+  }
```
ivan-kadykalo, line 17: Why not forEach?

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item) => obj[item[key]].push(item));
```
ivan-kadykalo, line 19: There is a way to make all these operations in only one loop instead of 3. Try to refactor your code.

ivan-kadykalo (changes requested):

PR #432, opened 2022-08-26
Task solution

sergii-nosachenko:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    const itemValue: string = String(item[key]);
+
+    if (!Object.prototype.hasOwnProperty.call(grouped, itemValue)) {
```
AlexanderKolomiiets, line 11: Why are you using so complicated check if a object's key exist? )
P.S. In this case, you can use `({})` instead `Object.Prototype` if you're calling standard method.

AlexanderKolomiiets (approved):
🛸🚀

src/groupByKey.ts
sergii-nosachenko: @AlexanderKolomiiets Thanks, replaced with simpler solution

AlinaMatuschak (approved):
🚀🚀

PR #434, opened 2022-08-26
Solution

Dmytro-Matsehora:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.forEach((el) => {
```
OlehMy, line 8: in my opinion, 'item' is better name for this var then 'el')))

OlehMy (approved):
Good solution)

AlinaMatuschak (approved):

PR #436, opened 2022-08-27
Develop

VolodymyrChuyko:


src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+type Item = object;
```
AndreyJigijalkin, line 5: This extra Type is unnecessary

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+
+export function groupByKey(items: Item[], key: keyof Item): GroupsMap<Item> {
```
AndreyJigijalkin, line 7:
```suggestion
export function groupByKey<Item>(items: Item[], key: keyof Item): GroupsMap<Item> {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+  const group: GroupsMap<Item> = {};
+
+  items.forEach((item) => {
+    const groupKey = item[key];
```
AndreyJigijalkin, line 11:
```suggestion
    const groupKey = String(item[key]);
```

So you don't need your `type Item`

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+
+    if (!Object.prototype.hasOwnProperty.call(group, groupKey)) {
```
AndreyJigijalkin, line 13: You can simplify this part:) 

```suggestion
    if (!group[groupKey]) {
```

AndreyJigijalkin (approved):
There are few small issues, looking good to me, but IDK if it is ok for mentors:)

src/groupByKey.ts
VolodymyrChuyko: this simple solution is possible but less reliable

src/groupByKey.ts
VolodymyrChuyko: There are several ways to deal with that type. I still have doubts about which is the best...

dellachaise (approved):
Well done 👍

PR #449, opened 2022-09-01
function groupByKey was created (ts)

VasylZinchenko:


David-Shahmatov (approved):
🦾 Great job)

src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedByKey:GroupsMap<T> = {};
+
+  items.map((item: T) => {
+    return (groupedByKey[`${item[key]}`] = groupedByKey[`${item[key]}`]
+    || []).push(item);
+  });
```
AlinaMatuschak, line 11: O no 🙈
Should use `map` when need to create a new array with new values from the current array. But you don't catch the new array and you update object that exists outside the map callback 🥲

AlinaMatuschak (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedByKey:GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    return (groupedByKey[`${item[key]}`] = groupedByKey[`${item[key]}`]
+    || []).push(item);
+  });
```
AlinaMatuschak, line 11: Why do you return something from the callback on forEach? For what? You need just update the groupedByKey, it's all

AlinaMatuschak (approved):

PR #459, opened 2022-09-04
GroupByKey solution maked

ibondrk:
Created solution to group object elements by main key.


roma-shafihullin (approved):
good work! 💪

src/groupByKey.test.ts
```diff
@@ -24,10 +24,10 @@ describe('Function "groupByKey"', () => {
     expect(Object.keys(groupedData))
       .toEqual(['22', '23', '24', '26']);
 
-    expect(groupedData[22].map((student) => `${student.name} ${student.age}`))
+    expect(groupedData[22].map((student: { name: any; age: any; }) => `${student.name} ${student.age}`))
```
AndriiRut, line 27: Try to avoid using the type `any`. If it is possible use `string` and `numbre` types.

AndriiRut: if you want to show that argument is not required use `?`
`interface Object { name?: string, age?: number }`
this sign means that the value is possibly undefined.

AndriiRut (changes requested):

src/groupByKey.test.ts
```diff
@@ -24,10 +24,10 @@ describe('Function "groupByKey"', () => {
     expect(Object.keys(groupedData))
       .toEqual(['22', '23', '24', '26']);
 
-    expect(groupedData[22].map((student) => `${student.name} ${student.age}`))
+    expect(groupedData[22].map((student: { name?: string, age?: number }) => `${student.name} ${student.age}`))
       .toEqual(['Willie 22', 'Glenn 22']);
 
-    expect(groupedData[26].map((student) => `${student.name} ${student.age}`))
+    expect(groupedData[26].map((student: { name: any; age: any; }) => `${student.name} ${student.age}`))
```
denys-danyliuk, line 30: avoid using type any

denys-danyliuk (approved):
good job👍
pay attention to the comment

PR #460, opened 2022-09-04
group_by_key function implemented

roma-shafihullin:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const obj: GroupsMap<T> = {};
```
ibondrk, line 6: This wont affect the functionality, but in general it's better to name variables in such a way that they describe what is stored in them. In this case i mean `obj`. In my opinion it would be better to name it like `groupedByKey` or `groupedObject` something like this)

ibondrk (approved):

PR #461, opened 2022-09-07
Dev

DominoSl:


src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+  items.forEach((item) => {
+    const groupKey = String(item[key]);
+
+    if (groupKey === undefined) {
+      return;
+    }
+
+    if (!groups[groupKey]) {
+      groups[groupKey] = items.filter((el) => el[key] === item[key]);
+    }
```
Bohdan-Maksymiuk, line 17: Now you iterate through item on every iteration. It is bad solution
you can:
* create an empy array as value of key
* push item into this array after check

Bohdan-Maksymiuk (changes requested):

qoa11a (approved):

PR #485, opened 2022-10-03
group funcrion has been created

illia-sosnytskyi:


artem-musii (approved):
GJ

Anna-Tkachenko (changes requested):
Try to solve it in easier way using one array iteration instead of two

Denchepornyuk (approved):

PR #486, opened 2022-10-03
solution

AntonVaida:


src/groupByKey.ts
```diff
@@ -1,7 +1,19 @@
...
+export function groupByKey<T>(items:T[], key:keyof T):GroupsMap<T> {
+  const result:GroupsMap<T> = {};
+
+  items.forEach((item:T) => {
+    result[String(item[key])] = [];
+  });
+
+  items.forEach((item:T) => {
+    result[String(item[key])].push(item);
+  });
```
grafit-off, line 15: It's not critical, but I think you need only one cycle for this, something like
```suggestion
  items.forEach((item: T) => {
    if (!result[String(item[key])]) {
      result[String(item[key])] = [];
    }

    result[String(item[key])].push(item);
});
```

grafit-off (approved):
GJ 🔥 Your solution is more readable than mine and I like it!

src/groupByKey.ts
Anna-Tkachenko: agree, you can do in one iteration

Anna-Tkachenko (approved):

PR #489, opened 2022-10-04
groupByKeys was implemented

ArturNagaycev:


Vladosik1337 (approved):
Nice work🔥

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    groupedItems[`${item[key]}`] = [];
+  });
+
+  items.forEach((item: T) => {
+    groupedItems[`${item[key]}`].push(item);
+  });
```
OleksiiBlezniuk, line 17: Let's make our logic simpler, why to iterate array twice if we can do it only once 🙂

OleksiiBlezniuk (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    if (!groupedItems[`${item[key]}`]) {
```
OleksiiBlezniuk, line 12: Or we can save this value to constant to avoid using interpolation and property value reading every time

OleksiiBlezniuk (approved):

PR #491, opened 2022-10-04
Group by key

Vladosik1337:


a-nahaitsev (approved):
Well done!

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+export function groupByKey<T>(items:T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
+    result[String(items[i][key])] = [];
+  }
+
+  for (let i = 0; i < items.length; i += 1) {
+    result[String(items[i][key])].push(items[i]);
+  }
```
OleksiiBlezniuk, line 14: Let's make our logic simpler, why to iterate array twice if we can do it only once? 🙂

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #500, opened 2022-10-05
solution was added

Maksym845:


varkovroma (approved):
well done!

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+  items.forEach((current) => {
+    const nKey = current[key] as keyof GroupsMap<T>;
+
+    if (result[nKey] === undefined) {
+      result[nKey] = [current];
+    } else {
+      result[nKey].push(current);
+    }
+  }, {});
```
Denchepornyuk, line 16: what is empty object doing here?

Denchepornyuk (changes requested):

src/groupByKey.ts
Maksym845: Oh, yeah I forgot about this object because firstly I wrote this by reduce method.

Denchepornyuk (approved):

PR #501, opened 2022-10-05
GroupByKey solution

Yevheniya-Deryka:


glugovsky (approved):
👍

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    if (`${item[key]}` in groupedItems) {
```
naziamoff, line 9: better to create separate variable for item[key], you use it 3 times

naziamoff: and interpolation looks quite strange here, think of a way to remove it

naziamoff (changes requested):

src/groupByKey.ts
Yevheniya-Deryka: Done!

naziamoff (approved):
Nice!

PR #503, opened 2022-10-05
Items have been grouped by key

iam-datsenko:


iPanda177 (approved):
GJ

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
+    if (!grouped[String(items[i][key])]) {
+      grouped[String(items[i][key])] = [];
+    }
+
+    grouped[String(items[i][key])].push(items[i]);
```
SerhiiKirik, line 13: You repeat this 3 times "String(items[i][key])"
You can create a variable for this, it will be more clean & readable
Also, you can do this with forEach or with for of, it's easier to read

silchyn (changes requested):
Good points above

silchyn (approved):
Cool!

PR #504, opened 2022-10-05
solution

nikitaovch99:


src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedObject: GroupsMap<T> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
+    if (!groupedObject[String(items[i][key])]) {
+      groupedObject[String(items[i][key])] = [];
+    }
+    groupedObject[String(items[i][key])].push(items[i]);
+  }
```
OleksiiBlezniuk, line 13: Ok, but would be nice to use array iteration methods instead of loop (reduce will be nice here)

OleksiiBlezniuk (approved):

PR #513, opened 2022-10-06
solution

myroslav-diiak:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const obj: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    if (`${item[key]}` in obj) {
```
anastasiia-nahaiska, line 9: In these tests it works, but if some key was named as a key in "Object.prototype" it wouldn't work. So maybe `hasOwnProperty` would be a better solution. But I'm not sure

anastasiia-nahaiska (approved):
GJ!

silchyn (changes requested):
Good point above

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const obj: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    if (obj[`${item[key]}`]) {
```
naziamoff, line 9: why not hasOwnProperty?

naziamoff (changes requested):

src/groupByKey.ts
myroslav-diiak: When i using hasOwnProperty linter throws an error

naziamoff (approved):
🔥

PR #528, opened 2022-10-19
First commit

Bibicutu:


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedByKey: GroupsMap<T> = {};
+  const keysInGrouped: any[] = [];
```
fatihsamur, line 7: using 'any' type might not be the best approach.

fatihsamur (approved):
check the comment

src/groupByKey.ts
VinnieJ-2k20: Yep, it's a bad idea. Don't use `any`

VinnieJ-2k20 (changes requested):

ihor-jpeg (approved):

PR #530, opened 2022-10-20
solution

Naveed-Shoukat:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+  for (let i = 0; i < items.length; i += 1) {
+    const resultKey = String(items[i][key]);
+
+    if (resultKey in result) {
+      result[resultKey].push(items[i]);
+    } else {
+      result[resultKey] = [items[i]];
+    }
+  }
```
kamal-tufekcic, line 16: Maybe use iteration methods?

VinnieJ-2k20: Would be great to use array methods.

VinnieJ-2k20 (approved):

PR #535, opened 2022-11-04
solution

serhii-yunnikov:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: object[], key: string): GroupsMap<object> {
+  const objectResult = {} as GroupsMap<object>;
```
Bohdan-Maksymiuk, line 6:
```suggestion
  const objectResult = {};
```
try to avoid [Type Assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) and `object` type
just accept type in function as a generic and use

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: object[], key: string): GroupsMap<object> {
```
Bohdan-Maksymiuk, line 5:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
```

Bohdan-Maksymiuk (changes requested):
Feel free to ask in `fe_chat` if smth is unclear

loralevitska (approved):
good job!

PR #537, opened 2022-11-05
Implement sortBy function

anastasiia-tilikina:


src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<R>(items: R[], key: keyof R): GroupsMap<R> {
+  const condition: R[keyof R][] = items.map((item: R) => item[key]);
+
+  const groups: GroupsMap<R> = {};
+
+  condition.forEach((el: R[keyof R]) => {
```
tetiana-ivanchenko, line 10: you can combine map and forEach to avoid one loop

src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
...
+    groups[el as string] = items.filter((item: R) => item[key] === el);
```
tetiana-ivanchenko, line 11: also you can resolve this task without using nesting loops,

actually you can resolve whole task using just one loop:

just create a new object, then in forEach, define what group (key) the element should have, and then if you don't already have that group in your new object, create a new array with that key, then push your element there.

If you have any questions - feel free to ask them to the fe_pt2 group!)

tetiana-ivanchenko (changes requested):

src/groupByKey.ts
anastasiia-tilikina: thanks for your advice)
didn't notice this way of implementation)

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<R>(items: R[], key: keyof R): GroupsMap<R> {
+  const groups: GroupsMap<R> = {};
+
+  items.forEach((item: R) => {
```
tetiana-ivanchenko, line 8:
```suggestion
  items.forEach((item) => {
```

it's redundand because TS already knows, that you items type - is array of "R" - so item type - is R

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+    const condition: R[keyof R] = item[key];
```
tetiana-ivanchenko, line 9:
```suggestion
    const condition: stirng = item[key];
```

we already knows that R[keyof R] - string

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+    if (!groups[condition as string]) {
+      groups[condition as string] = [];
+    }
+
+    groups[condition as string].push(item);
```
tetiana-ivanchenko, line 15:
```suggestion
    if (!groups[condition]) {
      groups[condition] = [];
    }

    groups[condition].push(item);
```

tetiana-ivanchenko (changes requested):

liliia-livandovskaya (approved):
Great job 🔥

PR #541, opened 2022-11-10
Develop

SergIats:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const itemKey = String([item[key]]);
+
+    const filtredItems = items.filter((elem) => String(elem[key]) === itemKey);
+
+    result[itemKey] = filtredItems;
```
tetiana-ivanchenko, line 13: why do you need filter here?
you just need to check, if this itemKey property is already exist in you result, just push current item into it, or if it diesn't exist, create new array, and put your item into it

tetiana-ivanchenko (changes requested):

SergIats:
Hi, Tetiana! Thank you for the review, I have implemented your recommendation)
But I'm not sure that I clearly understood why this way better? Because "filter" is more difficult process for processor?

nazarmatsevych (approved):

PR #542, opened 2022-11-10
Solution

Illia-Skladnik:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: object[], key: string):
+GroupsMap<{[key: string]: object[]}> {
+  const groupedObject: GroupsMap<{[key: string]: object[]}> = {};
```
iryna-lelyk, line 7:
```suggestion
GroupsMap<object> {
  const groupedObject: GroupsMap<{[key: string]: object[]}> = {};
```
You can use it like that. In type GroupsMap already mentioned the structure of object with keys and values like that.

iryna-lelyk (approved):
Good job!

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: object[], key: string):
```
denys-danyliuk, line 5: use generic and keyof: https://mate-academy.github.io/fe-program/js/extra/typescript

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+GroupsMap<{[key: string]: object[]}> {
```
denys-danyliuk, line 6: use generic instead

src/groupByKey.ts
denys-danyliuk: no, just use generic

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #543, opened 2022-11-11
Group by key was added

iryna-lelyk:


Illia-Skladnik (approved):
Looks good!

denys-danyliuk (changes requested):
find a solution using generics

PR #545, opened 2022-11-11
solution

IOneStarI:


src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<object> {
```
karyna-salnykova, line 5: I think its type will be more clear if you write not just string but extend of T

src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
...
+  return items.reduce((hash, obj) => {
```
karyna-salnykova, line 6: Wow, I had no idea how to use reduce in this task! Cool

src/groupByKey.ts
mykola-koval: also don't use <object>

src/groupByKey.ts
mykola-koval: why not forEach?

mykola-koval (changes requested):

src/groupByKey.ts
IOneStarI: just came to my head.  I will make with forEach)

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grougedItems: GroupsMap<T> = {};
```
mykola-koval, line 6:
```suggestion
  const groupedItems: GroupsMap<T> = {};
```

mykola-koval (approved):

PR #549, opened 2022-11-12
implemented groupByKey function

Artem66:


src/groupByKey.test.ts
```diff
@@ -168,113 +168,113 @@ describe('Function "groupByKey"', () => {
...
-        language: 'German',
-        pages: 720,
-        title: 'The Magic Mountain',
-        year: 1924,
-      }],
-    });
-  });
+  // test('should group books by language', () => {
+  //   expect(groupByKey(books, 'language')).toEqual({
+  //     Japanese: [{
```
masha-girya, line 173: just interesting, why did you commented it out?)

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items:[], key: string): GroupsMap<object> {
+  const result: GroupsMap<object> = {};
```
masha-girya, line 6: i think that GroupsMap should be with <T> generic like it was in condition. Also, key in parameters should be not a string type, but 'keyof T'

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items:[], key: string): GroupsMap<object> {
```
masha-girya, line 5: In general function has to take <T> generic, but your function do not. And this <T> is actually an array of items (items: T[ ])

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+  const result: GroupsMap<object> = {};
+
+  items.forEach((item: {[key: string]: any}): void => {
```
masha-girya, line 8: maybe it can work without 'void'

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+    if (!Object.prototype.hasOwnProperty.call(result, item[key])) {
+      result[item[key]] = [item];
+    } else {
+      result[item[key]].push(item);
```
masha-girya, line 12: There you could check if this key is already in object by using !result[key]

masha-girya (changes requested):
i think, it is a bit complicated. If you will use generic <T> in this function it will be more simple. I recommend you to rewatch TASK#3 in the last lecture, there is an answer how to work with kyes, which type is uncknown for TS

masha-girya (approved):
good job!)

src/groupByKey.ts
```diff
@@ -1,7 +1,19 @@
...
+  const gropedByKey: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const currrentKey = String(item[key]);
+
+    if (!gropedByKey[currrentKey]) {
+      gropedByKey[currrentKey] = [item];
+    } else {
+      gropedByKey[currrentKey].push(item);
+    }
```
illiarerun, line 15: you can invert your condition to just **add blank array**, and **then** push your item to get rid of `else`

src/groupByKey.ts
```diff
@@ -1,7 +1,19 @@
...
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const gropedByKey: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const currrentKey = String(item[key]);
```
illiarerun, line 9:
```suggestion
    const currentKey = String(item[key]);
```

illiarerun (changes requested):

src/groupByKey.ts
illiarerun: pay attention to comments, fix it

src/groupByKey.ts
```diff
@@ -1,7 +1,21 @@
-type GroupsMap<T> = {
+type GroupsMap <T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const gropedByKey: GroupsMap<T> = {};
```
illiarerun, line 6:
```suggestion
  const groupedByKey: GroupsMap<T> = {};
```

src/groupByKey.ts
```diff
@@ -1,7 +1,21 @@
...
+  items.forEach((item) => {
+    const currrentKey = String(item[key]);
+
+    if (gropedByKey[currrentKey]) {
+      return gropedByKey[currrentKey].push(item);
+    }
+
+    gropedByKey[currrentKey] = [];
+
+    return gropedByKey[currrentKey].push(item);
```
illiarerun, line 17: that's too much, also `forEach` doesn't return anything, invert your solution as I said
```suggestion
    if (!groupedByKey[currentKey]) {
      here add blank array
    }

    here push
```

illiarerun (changes requested):

illiarerun (approved):

PR #551, opened 2022-11-12
grouped by key MG

masha-girya:


src/groupByKey.ts
```diff
@@ -1,7 +1,24 @@
+
 type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  let obj: GroupsMap<T> = {};
```
Artem66, line 7: I think you can change name something like grouped or so on. And what about let?

Artem66 (approved):

src/groupByKey.ts
masha-girya: oops, you are right!) let because in forEach method I change this object

src/groupByKey.ts
```diff
@@ -1,7 +1,24 @@
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  let groupedByKey: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    const value = item[key];
+
+    if (typeof value === 'string' || typeof value === 'number') {
```
mykola-koval, line 12: why do u need this check? maybe let's find a simpler solution

mykola-koval (approved):

PR #552, opened 2022-11-12
solution

Anna-Shapovalova:


Svitlana-Yudina (approved):
🔥 it's great!)

src/groupByKey.ts
```diff
@@ -2,6 +2,11 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedObject = items.reduce((acc, current) => ({
+    ...acc,
+    [String(current[key])]: items.filter((item) => item[key] === current[key]),
+  }), {});
```
illiarerun, line 9: 🔥

illiarerun, line 9: it'd be better to explicitly add a type, like this:
```suggestion
  const groupedObject = items.reduce((acc: GroupsMap<T>, current) => ({
    ...acc,
    [String(current[key])]: items.filter((item) => item[key] === current[key]),
  }), {});
```

or like this:
```suggestion
  const groupedObject = items.reduce((acc, current) => ({
    ...acc,
    [String(current[key])]: items.filter((item) => item[key] === current[key]),
  }), {} as GroupsMap<T>);
```

illiarerun (approved):

src/groupByKey.ts
Anna-Shapovalova: I got U 👍

PR #555, opened 2022-11-12
Solution

taraselnik:


Anna-Shapovalova (approved):
Nice code)

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+  const groupedObj: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const itemKey = `${item[key]}`;
+
+    if (groupedObj[itemKey]) {
+      groupedObj[itemKey].push(item);
+    } else {
+      groupedObj[itemKey] = [item];
+    }
```
illiarerun, line 15: you can invert your condition to just add blank array, and then push your item to get rid of `else`

illiarerun (changes requested):

src/groupByKey.ts
taraselnik: ok

illiarerun (approved):

PR #559, opened 2022-11-13
Solution

referee1988:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: Array<T>, key: keyof T): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    const itemKey = String(item[key]);
+
+    if (!groupedItems[itemKey]) {
+      groupedItems[itemKey] = [item];
+    } else {
```
mitattt, line 13: You could not use this else. Just write it down below without else.

mitattt (approved):
Good job

src/groupByKey.ts
mykola-koval: yep, you don't need else here

mykola-koval (approved):

PR #581, opened 2022-11-28
solution

olenitut:


src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(
+  items: object[], key: keyof object,
+): GroupsMap<object> {
+  const result: GroupsMap<object> = {};
+
+  items.forEach((item: object) => {
+    const keyValue: any = item[key];
```
Viktor-Kost, line 11: Try to avoid using any as a type

```suggestion
    const keyValue = item[key] as string;
```

Viktor-Kost (changes requested):
Well done!

A little change needed

Viktor-Kost (approved):
Nice!)

PR #585, opened 2022-12-02
ts_group_by_key

kolya-movchan:


src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
...
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T, K extends keyof T>(items: T[], key: K)
+  : GroupsMap<T> {
+  return items.reduce((prev, current) => {
+    const thisKey = [current[key]];
+
+    return {
+      ...prev,
+      [thisKey as any]: items.filter((object) => object[key] === current[key]),
```
AlisaVlasova, line 12: should thisKey have any type?

AlisaVlasova (changes requested):

src/groupByKey.ts
kolya-movchan: @AlisaVlasova TS shows an error then
image
image

AlisaVlasova (approved):

PR #586, opened 2022-12-02
Solution

Julia2063:


src/groupByKey.ts
```diff
@@ -1,7 +1,17 @@
+/* eslint-disable array-callback-return */
```
Anastasiia-Svintsova, line 1: Don't disable eslint. Try to solve the error without it

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (approved):
Good job!

PR #592, opened 2022-12-14
solution

nikachu404:


AntonMartynes (approved):
Good

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T) : GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyValue: string = `${item[key]}`;
```
solaryasha, line 9:
```suggestion
    const keyValue = `${item[key]}`;
```

solaryasha: Also, it's better to use `String` constructor for type conversion

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T) : GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
```
solaryasha, line 8: What about the `reduce`? You could try it, cause reduce method also generic :)

solaryasha (approved):

PR #598, opened 2022-12-16
solution

korchovskyi:


src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((acc, item) => {
+    const value = item[key];
```
APetrinko, line 7:
```suggestion
    const value = String(item[key]);
```

src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
...
+
+    acc[value as string] = acc[value as string] || [];
+    acc[value as string].push(item);
```
APetrinko, line 10: DRY

```suggestion
    acc[value] = acc[value] || [];
    acc[value].push(item);
```

APetrinko (changes requested):
everything is fine except for one point 😉

PR #600, opened 2022-12-19
Solution

OlggaJul:


superpooperxxx (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedByKey = {};
+
+  items.forEach((item) => {
```
solaryasha, line 8: What about the `reduce`? You could try it, cause reduce method also generic :)

solaryasha (approved):

src/groupByKey.ts
OlggaJul: 👍🏻

PR #604, opened 2022-12-19
solution

AntonMartynes:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouppedObject: GroupsMap<T> = {};
```
solaryasha, line 6: Avoid using data structures in the variable names. It's really confusing

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item) => {
+    const val = String(item[key]);
```
solaryasha, line 9: What is `val` ?
[This](https://val.ua/) ?
Or [this](https://uk.wikipedia.org/wiki/VAL_(%D0%B3%D1%83%D1%80%D1%82)) ?

solaryasha (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedObj: GroupsMap<T> = {};
```
solaryasha, line 6: `obj` is still the data structure :)
```suggestion
  const groupedItems: GroupsMap<T> = {};
```

solaryasha (approved):

src/groupByKey.ts
AntonMartynes: Thank You !)

PR #605, opened 2022-12-19
Solution

nurdanokcu:


UmizDemud (approved):
👍 well done

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<R>(
+  items: R[],
+  key: keyof R,
+): GroupsMap<R> {
+  return items.reduce((acc: GroupsMap<R>, item: R) => {
+    const groupKey:string = `${item[key]}`;
```
meforoma, line 10: type will be automatically assigned to string if the right-hand-side after `=` sign is a string / string literal
```suggestion
    const groupKey = `${item[key]}`;
```

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+    const groups = { ...acc };
+
+    if (groups[groupKey]) {
+      groups[groupKey].push(item);
+    } else {
+      groups[groupKey] = [item];
+    }
```
meforoma, line 17: i believe you could try something like this:
```suggestion
      groups[groupKey] = (groups[groupKey] || []).concat(item);
```

meforoma (approved):
✅ 
checkout few comments

PR #609, opened 2022-12-19
Solution

batstolya:


Anastasia4sik (approved):
Looks good

src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T):GroupsMap<T> {
+  const groupByObject:GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const myKey = String(item[key]);
+
+    groupByObject[myKey] = items.filter((person) => item[key] === person[key]);
```
ivan-kadykalo, line 11: In this case, you will reassign your key a couple of times with the same values.
Let's check if that key exist and push values to the according key

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #611, opened 2022-12-19
solution

lefteris-L:


src/groupByKey.ts
```diff
@@ -1,7 +1,24 @@
-type GroupsMap<T> = {
+export interface GroupsMap<T> {
```
UmizDemud, line 1: I like to see this style in JS, didn't know we could do that. Reminds me of C langs

UmizDemud (commented):
👍 
cool

UmizDemud (approved):

src/groupByKey.ts
```diff
@@ -1,7 +1,24 @@
...
+      newKeys.push(item[key]);
+    }
+  });
 
-export function groupByKey(items, key) {
-  // write code here;
+  return newKeys.reduce((prev: GroupsMap<Item>, newKey: KeyTypes) => ({
+    ...prev,
+    [newKey]: items.filter((item: Item) => item[key] === newKey),
+  }), {});
```
meforoma, line 23: i'm pretty sure you could have it done with just one `.reduce`
try this out and share any concerns.
```
items.reduce((acc, item) => {
    acc[key] = (...acc[key] || []).concat(item)

    return acc;
  }, {})
```

meforoma (changes requested):
please try out a more graceful approach - see the comment

src/groupByKey.ts
lefteris-L: Excellent. Thanks 👍

meforoma (approved):

PR #618, opened 2022-12-20
Solution

Yuriy-Khoptyanii:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+
+  const result: GroupsMap<T> = items.reduce((prev, value) => {
+    const groupKey = String(value[key]);
+    const copy = { ...prev };
+
+    if (!(groupKey in prev)) {
+      copy[groupKey] = [value];
+    } else {
+      copy[groupKey].push(value);
+    }
```
APetrinko, line 16: it makes more sense to me
```suggestion
    if (groupKey in prev) {
      copy[groupKey].push(value);
    } else {
      copy[groupKey] = [value];
    }
```

APetrinko (changes requested):
everything is ok except for one little moment 😉

src/groupByKey.ts
Yuriy-Khoptyanii: from permutation - the amount does not change.
I think it is not important to change.
let's leave it to the discretion of the mentor)

APetrinko (approved):

valerii-derkach (approved):

PR #629, opened 2022-12-20
Solution

TetKopot:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const sorted = String(item[key]);
```
Anastasia4sik, line 9: I don't know if it is necessary, maybe you need to add a type for the argument item and for sorted

Anastasia4sik (approved):
Looks good, but there is one advice

src/groupByKey.ts
TetKopot: VSCode doesn't know about that problem)

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
```
ivan-kadykalo, line 6: Grouped what?

```suggestion
  const groupedKeys: GroupsMap<T> = {};
```

src/groupByKey.ts
ivan-kadykalo, line 9: Sorted what?

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #636, opened 2022-12-21
added groupByKey function

TheOneThatKnowsNothing:


src/groupByKey.ts
```diff
@@ -2,6 +2,27 @@ type GroupsMap<T> = {
...
+
+  const maps
+    = groupKeys
+      .map((groupKey) => {
+        const map: GroupsMap<I>
+          = {
+            [`${groupKey}`]: items.filter((item : I) => item[key] === groupKey),
+          };
+
+        return map;
```
Mykhailyk-Vadym, line 20: maybe we can just return map without creating const?

Mykhailyk-Vadym (approved):
Looks good and interesting

src/groupByKey.ts
```diff
@@ -2,6 +2,27 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<I>(items: I[], key: keyof I): object {
+  let groupKeys = items.map((item: I) => item[key]);
+
+  groupKeys
```
solaryasha, line 8: Avoid variable's reassigning

src/groupByKey.ts
solaryasha: ++++

solaryasha (changes requested):

Ihor-Karpyn (approved):

PR #640, opened 2022-12-21
solution

olena-yanovska:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((prev, current) => {
+    const groupKey = current[key];
```
APetrinko, line 7:
```suggestion
    const groupKey = String(current[key]);
```

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+    const result = { ...prev };
+
+    if (!result[groupKey as string]) {
+      result[groupKey as string] = [];
+    }
+
+    result[groupKey as string].push(current);
```
APetrinko, line 14: DRY
```suggestion
    if (!result[groupKey]) {
      result[groupKey] = [];
    }

    result[groupKey].push(current);
```

APetrinko (approved):
U almost there 😉

APetrinko (changes requested):
U almost there 😉

src/groupByKey.ts
olena-yanovska: Thanks!)

APetrinko (approved):
🔥

Ihor-Karpyn (approved):

PR #646, opened 2022-12-22
Group by key task

Mohameddahi:


naziamoff (approved):
great usage of reduce method 🔥

PR #648, opened 2022-12-23
sol1

dolnaro:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+  const groupItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const sortValue = String(item[key]);
+
+    if (!groupItems[sortValue]) {
+      groupItems[sortValue] = [];
+    }
+    groupItems[sortValue].push(item);
+  });
```
naziamoff, line 15: would be good if done with `.reduce`

naziamoff (approved):

PR #650, opened 2022-12-26
Solution

ShkarupaDenys:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((accum: GroupsMap<T>, item) => {
+    const itemKey = item[key];
+    const result = accum;
```
TarasHoliuk, line 8: you can avoid creating `result` variable in this case, and do manipulation with `accum` (you don't reassign value entirely like this `accum = { someProperty: someValue }`)

TarasHoliuk (changes requested):

TarasHoliuk (approved):

PR #653, opened 2022-12-28
Group by key

maksym-kryvolap:


src/groupByKey.ts
```diff
@@ -1,7 +1,20 @@
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
+    const value = String(items[i][key]);
+
+    if (Object.prototype.hasOwnProperty.call(result, value)) {
```
TarasHoliuk, line 12: why do you use `call` here instead of doing it in this (shorter and clearer) way: `result.hasOwnProperty(value)`?

TarasHoliuk (changes requested):

src/groupByKey.ts
maksym-kryvolap: Because, the linter swears and writes that there is no way to use such a call

TarasHoliuk (approved):

PR #658, opened 2023-01-09
ts_group_by_key

dima-korovai:


src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.map((item) => {
+    const x = String(item[key]);
```
polosanya, line 9: Let's use more meaningful names than `x` and `result`. It will help to make your code cleaner and more understandable.

polosanya (changes requested):
Almost perfect!

src/groupByKey.ts
dima-korovai: Thank you!

polosanya (approved):
Good job!

PR #660, opened 2023-01-10
Solution

tarasavage:


IvanFesenko (approved):
It's ok. For practice, try to resolve this task using Array.reduce()

PR #662, opened 2023-01-11
Solution

Iksich:


naziamoff (approved):
good job. Could be implemented with `.reduce` method

PR #664, opened 2023-01-14
Task was done

Sergey-Mironenko:


src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
...
+  const result: GroupsMap<Group> = {};
+
+  items.map((el): void => {
+    if (!(el[key] in result)) {
+      result[`${el[key]}`] = [];
+    }
+
+    result[`${el[key]}`].push(el);
+
+    return undefined;
```
TarasHoliuk, line 19: remove it and use `forEach` instead of `map` to iterate through an array

src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+type Group = {
+  [key: string]: number | string;
+};
+
+export function groupByKey(items: Group[], key: string): GroupsMap<Group> {
```
TarasHoliuk, line 9: let's use more TS features

use generic type in function
change types for function arguments (you don't need your type `Group`, use `keyof` for `key`)

```suggestion
export function groupByKey<T>(items: Group[], key: string): GroupsMap<Group> {
```

TarasHoliuk (changes requested):

vlad-demchuk (approved):

PR #669, opened 2023-01-17
Add groupByKey function

mykyta-krainik:


src/groupByKey.ts
```diff
@@ -1,7 +1,19 @@
-type GroupsMap<T> = {
+interface GroupsMap<T> {
```
Anastasiia-Svintsova, line 1: BTW type is also ok

Anastasiia-Svintsova (approved):
👍

PR #670, opened 2023-01-18
solution

Avichez:


OlhaYevstifieieva (approved):
Clever solution with 'Object.prototype' 👋🏻

Anna-Tkachenko (approved):
⭐️

PR #678, opened 2023-01-18
solution

AnnyKoval:


AnnyKoval:
type GroupsMap<T> = {
  [key: string]: T[];
};

export function groupByKey <T>(items: T[], key: keyof T): GroupsMap<T> {
  const obj: GroupsMap<T> = {};

  items.forEach((element) => {
    const filterKey = String(element[key]);

    if (!obj[filterKey]) {
      obj[filterKey] = [];
    }
    obj[filterKey].push(element);
  });

  return obj;
}

IvanFesenko (approved):
it's ok, Try implements same using Array.reduce

PR #679, opened 2023-01-19
Solution

max-rozzhalovets:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+      item[key] === itemForFilter[key]
+    ));
+
+    return acc[item[key]]
+      ? acc
+      : {
+        ...acc,
+        [item[key]]: add,
+      };
+  }, {});
```
NVyrsta, line 18: As I see, you are also a fan of the reduce method, like me ;0)
here is another possible solution:
`
  return items.reduce((acc: GroupsMap<T>, item) => {
    const groupBy = `${item[key]}`;

    acc[groupBy] = (acc[groupBy] || []).concat(item);

    return acc;
  }, {});
}
`

NVyrsta (approved):
Well done ༼ つ ◕_◕ ༽つ

Denchepornyuk (approved):

PR #680, opened 2023-01-19
task done

Bogdan-Kutasevich:


vanst1 (approved):
Good job)

src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
...
+    result[groupKey] = [];
+  });
+
+  items.forEach((item) => {
+    const groupKey = String(item[key]);
+
+    if (Object.hasOwnProperty.call(result, groupKey)) {
+      result[groupKey].push(item);
+    }
+  });
```
Denchepornyuk, line 20: you can do all logic in one forEach method

Denchepornyuk (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T):GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const groupKey = String(item[key]);
+
+    if (!result[groupKey]) {
+      result[groupKey] = items.filter((elem) => groupKey === String(elem[key]));
```
Anna-Tkachenko, line 12: do you need to call filter method inside the forEach one?
you've already have a cycle, where you go through all of elements
just check each of them one by one, and add to the needed key

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #685, opened 2023-01-19
solution

BrandonSunrise:


src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+  const findKeys: Set<any> = new Set(items.map((item: T) => item[key]));
```
anna-inozemets, line 7: Maybe it is possible to change value "any" here

anna-inozemets (approved):
Pretty good job!🐇

src/groupByKey.ts
ArturErmolenko: agree , we doesn't use `any`, we want our code will be predictable

ArturErmolenko (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #691, opened 2023-01-21
Solution

OleksandrOse:


src/groupByKey.ts
```diff
@@ -2,6 +2,9 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((prev, user: T) => ({
+    ...prev,
+    [user[key] as string]: items.filter((val: T) => val[key] === user[key]),
+  }), {} as GroupsMap<T>);
```
SanyaBratashchuk, line 9: use generic for reduce instead of type cast

SanyaBratashchuk (changes requested):

Anastasiia-Svintsova (approved):

PR #699, opened 2023-01-28
ts / group by key

valntyn:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((acc: GroupsMap<T>, el) => {
+    const mainKey = String(el[key]);
+
+    if (acc[mainKey] === undefined) {
```
VitaliyBondarenko1982, line 9:
```suggestion
    if (!acc[mainKey] ) {
```

VitaliyBondarenko1982 (changes requested):

src/groupByKey.ts
Volodymyr-Melnyk-Fron10der: The problem is still here

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Check you really pushed you changes into remote branch

VitaliyBondarenko1982 (approved):

PR #700, opened 2023-01-28
All tests is ok

diolna:
All tests are ok. But I could not exclude the use of type casting.


src/groupByKey.ts
```diff
@@ -2,6 +2,24 @@ type GroupsMap<T> = {
...
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T, K extends keyof T>(items: T[], key:K)
+  : GroupsMap<T> {
+  type GroupsMapKey = keyof GroupsMap<T>;
+
+  const result = items.reduce((prev, item) => {
+    const valueOfKey = <GroupsMapKey>item[key];
+    const itemsByValueKey = items.filter((filterItem) => {
+      return filterItem[key] === item[key];
```
AlisaVlasova, line 12:
```suggestion
      return filterItem[key as keyof T] === item[key as keyof T]);
```

AlisaVlasova: try this

AlisaVlasova (changes requested):

diolna:
Fixed

PR #703, opened 2023-01-29
 ts_group_by_key

07Spirit:


src/groupByKey.ts
```diff
@@ -1,7 +1,18 @@
...
-  [key: string]: T[];
-};
-
-export function groupByKey(items, key) {
+export function groupByKey<T>(items: T[], key: keyof T): object {
   // write code here;
+  const result: any = {};
+
+  items.forEach((a: T) => {
+    const variable: any = a[key];
```
polosanya, line 6: Same here. Also, try not to use any. Set the correct type instead

src/groupByKey.ts
```diff
@@ -1,7 +1,18 @@
...
+
+    if (Object.keys(result).includes(String(variable))) {
```
polosanya, line 8: You can check it using `in` or `hasOwnProperty`

polosanya (changes requested):
Almost done!

src/groupByKey.ts
07Spirit: In this place, "Result" as object, adding inside any type of element(number, string etc) example sort by age:
{
...
24: [
  {
      "name": "Diana",
      "surname": "Dorsey",
      "age": 24,
      "married": false,
      "grades": [3, 3, 4, 5, 4, 3, 5, 5]
    },
    {
      "name": "Pamela",
      "surname": "Casillas",
      "age": 24,
      "married": false,
      "grades": [4, 5, 4, 5, 5, 4, 3, 2, 3, 3, 3, 2]
    }
],
...
},
"Variable" can have any names taken from element of object[key]:
"Diana" or 24 or [3, 3, 4, 5, 4, 3, 5, 5]

In condition (ReadME), I dont see limit of type for result.

src/groupByKey.ts
07Spirit: Thanks)

src/groupByKey.ts
07Spirit: Remake for string.

src/groupByKey.ts
```diff
@@ -1,7 +1,18 @@
-type GroupsMap<T> = {
-  [key: string]: T[];
-};
-
-export function groupByKey(items, key) {
+export function groupByKey<T>(items: T[], key: keyof T): object {
   // write code here;
+  const result: any = {};
```
polosanya, line 3:
```suggestion
  const result: GroupsMap<T> = {};
```
You had a type for it above. Use it
Also, it's better to use more descriptive names than just result

src/groupByKey.ts
```diff
@@ -1,7 +1,18 @@
...
+
+  items.forEach((itemsElement: T) => {
+    const variable: string = String(itemsElement[key]);
+
+    if (Object.prototype.hasOwnProperty.call(result, variable)) {
```
polosanya, line 8:
```suggestion
    if (result.hasOwnProperty(variable)) {
```

src/groupByKey.ts
```diff
@@ -1,7 +1,18 @@
...
-  [key: string]: T[];
-};
-
-export function groupByKey(items, key) {
+export function groupByKey<T>(items: T[], key: keyof T): object {
   // write code here;
+  const result: any = {};
+
+  items.forEach((itemsElement: T) => {
+    const variable: string = String(itemsElement[key]);
```
polosanya, line 6: Same with variable. Better specify what variable it is

polosanya (changes requested):
Let's polish it up)

src/groupByKey.ts
07Spirit: If I use code like this, I got error:
  12:16  error  Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins

src/groupByKey.ts
07Spirit: Its works! Thanks.

src/groupByKey.ts
07Spirit: In result, this variable will be string.

src/groupByKey.ts
```diff
@@ -1,7 +1,22 @@
...
+  [keyName: string]: T[];
 };
 
-export function groupByKey(items, key) {
+export function groupByKey<T>(items: T[], key: keyof T): object {
   // write code here;
+  const result: GroupsMap<T> = {};
+
+  items.forEach((itemsElement: T) => {
+    const variable: string = String(itemsElement[key]);
```
qoa11a, line 10: it's not about the type of variable but about its name. You shouldn't name variables "variable". It doesn't give any new information. Name it according to what it stores

qoa11a (changes requested):

src/groupByKey.ts
07Spirit: A, okay)

src/groupByKey.ts
07Spirit: I rename it)

Anastasiia-Svintsova (approved):

PR #704, opened 2023-01-29
solution

Deniskakaka:


src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey
+<T>(items: T[], key: keyof T): GroupsMap<T> {
+  // i know it is not good, but what type it will be?
+  const result: { [key: string]: T[keyof T] } = {};
+
```
IvanFesenko, line 9: You lready have this type - use it
```suggestion
  const result: GroupsMap<T> = {};

```

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+  for (let i: number = 0; i < items.length; i += 1) {
+    if (typeof result[items[i][key]] === 'undefined') {
+      result[items[i][key]] = [];
+    }
+
```
IvanFesenko, line 14: Better will be Array.forEach 
No need to check on typeof 
`if (!result[items[i][key]])`

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+    if (items[i][key] as unknown as 'string' in result) {
+      result[items[i][key]].push(items[i]);
+    }
+  }
```
IvanFesenko, line 18: Why type reseting? 
const groupKey = items[i][key] as string

IvanFesenko (changes requested):
Hey, check my comments

src/groupByKey.ts
Viktor-Kost: is not fixed

witflash: @Deniskakaka Does it not work if you use `GroupsMap<T>` here?

witflash (approved):

src/groupByKey.ts
Deniskakaka: Both options work, according to the tests, but ts swear at both

PR #709, opened 2023-02-03
object sorter

Mykola-Busarhin:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: [], key: keyof T): GroupsMap<T> {
+  const sortedObject: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    if (!sortedObject[item[key]]) {
```
ivbutsykin, line 9: You can move `item[key]` in separate variable (`value` for ex.) and when check as:
```
 if (!(value in sortedObject)) {
   ...
 }
 ```
 
 It's can improve readability

ivbutsykin (approved):

PR #712, opened 2023-02-07
Solution

Vlad-Fedorishchev:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const resultKey = <string>item[key];
```
agniya-i, line 9: you can use `String()` to type сonversion

agniya-i (approved):

PR #718, opened 2023-02-13
commit

skomskiy:


slavanosatskyi (approved):
good work! as an additional exercise you can try to solve this task with reduce method

PR #719, opened 2023-02-13
solution

vladyslav-krainyk:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+    const keyOfItem = String(item[key]);
+
+    if (!result[keyOfItem]) {
+      result[keyOfItem] = [];
+    }
+
+    result[keyOfItem].push(item);
+  });
+
+  return result;
```
etojeDenys, line 18: you can rewrite it using the Array.map() method

etojeDenys (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+    if (keyOfItem in result) {
+      result[keyOfItem].push(item);
+    } else {
+      result[keyOfItem] = [item];
+    }
+
+    return null;
+  });
+
+  return result;
```
tetiana-ivanchenko, line 20: why do you need to push in the new array, if map returns new array?

tetiana-ivanchenko (changes requested):

SanyaBratashchuk (approved):

PR #720, opened 2023-02-13
Solution

Sergey-Kovyazin:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+  const resultKeys = items.map((item) => item[key]);
+
+  resultKeys.forEach((resultKey) => {
+    if (String(resultKey)) {
+      result[String(resultKey)] = items
+        .filter((item) => item[key] === resultKey);
+    }
+  });
```
tetiana-ivanchenko, line 14: try to avoid this filter, because you do it multiple times for each case

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #731, opened 2023-02-22
initial commit

ihorivna007:


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey(
+  items: object[],
+  key: keyof object,
+): GroupsMap<object> {
+  const result: GroupsMap<object> = {};
+
+  items.filter((item: object) => {
+    if (!result[item[key]]) {
+      result[item[key]] = [];
```
mykhalenych, line 13: it isn't correct to change another array from filter, better to use forEach

mykhalenych (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(
+  items: object[],
+  key: keyof object,
```
DarkMistyRoom, line 7:
```suggestion
  key: string,
```

DarkMistyRoom: object key is always string

DarkMistyRoom (approved):

PR #732, opened 2023-02-23
solution

gabriella135:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const object: GroupsMap<T> = {};
+
+  items.forEach((e) => {
```
slavanosatskyi, line 8: e is not the best name here, just item would work fine

slavanosatskyi (approved):
good 👍 just one thing to consider, though it is a minor comment

PR #733, opened 2023-02-24
solution

ivanm2706:


src/groupByKey.ts
```diff
@@ -2,6 +2,24 @@ type GroupsMap<T> = {
...
+    if (!acc.includes(valueKey)) {
+      acc.push(valueKey);
+    }
+
+    return acc;
+  }, []);
+
+  keyArray.forEach((e: string) => {
+    obj[e] = items.filter((curr: object) => curr[key] === e);
+  });
```
IvanFesenko, line 22: You can do it just with one reduce just give object as initial value instead of array

IvanFesenko (changes requested):
Refactor without additional iterations

IvanFesenko (approved):
Well done

PR #734, opened 2023-02-24
Task was solved

Vasyl-Pavlenko:


IvanFesenko (approved):
It works, but better do it with `Array,reduce()`

PR #738, opened 2023-02-28
solution

Maks-Tsarenko:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const value = String(item[key]);
+
+    if (!(value in result)) {
+      result[value] = [];
+    }
```
SerhiiKirik, line 13: You can check like that or use hasOwnProperty
```suggestion
    if (!value[value])) {
      result[value] = [];
    }
```

SerhiiKirik (approved):
Well done 🚀

PR #740, opened 2023-03-01
Solution for Group By Key

ctk12:


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: any[], key: string): any {
+  const groupedObj: GroupsMap<any> = {};
+
+  items.map((item) => {
```
polosanya, line 8:
```suggestion
  items.forEach((item) => {
```
It is better to use forEach method, if you want only iterate through some array

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+    if (groupedObj[item[key]] === undefined) {
```
polosanya, line 9: It is better to check if property is present in object using "[in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in)" or "[hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty)"

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+      groupedObj[item[key]] = [];
+      groupedObj[item[key]].push(item);
+    } else {
+      groupedObj[item[key]].push(item);
```
polosanya, line 13: It would be great to create a variable for item[key] as it duplicates a lot

polosanya (changes requested):
Let's improve your code a bit)

qoa11a (approved):

PR #745, opened 2023-03-04
Group by key

Igor-Yaremchuk:


src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const itemValue = String(item[key]);
+
+    groupItems[itemValue] = items.filter((obj) => obj[key] === item[key]);
```
polosanya, line 11: Do you really need to use one more iterative method here?
I think you can solve it without nesting them.

polosanya (approved):
Good job! I'm approving it, but you can try to optimize your solution a bit)

PR #747, opened 2023-03-07
solution for groupByKey

vladyslav-mosor:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedObj: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const value = String(item[key]);
+
+    if (!(value in groupedObj)) {
+      groupedObj[value] = [];
+    }
```
Zelinskyi-Serhii, line 13: I heard that better use method hasOwnProperty.

Zelinskyi-Serhii (approved):
gj

src/groupByKey.ts
sergii-nosachenko: Pro hint: [Use OR assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment) instead of `if`.

sergii-nosachenko (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedObj: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const value = String(item[key]);
+
+    groupedObj[value] ||= [];
```
sergii-nosachenko, line 11:
```suggestion
    groupedObj[value] ??= [];
```
Tip: Here is an even better decision to prevent falsy values to be overwritten.

sergii-nosachenko (approved):
Good work 👍

PR #749, opened 2023-03-07
groupByKey solution

kateryna-babina:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+export function groupByKey<T>(items: T[], key: keyof T) : GroupsMap<T> {
+  return items.reduce((map:GroupsMap<T>, item: T) => {
+    const groupField = item[key] as string;
+    const groupObjectCopy: GroupsMap<T> = { ...map };
+
+    if (Object.prototype.hasOwnProperty.call(groupObjectCopy, groupField)) {
+      groupObjectCopy[groupField].push(item);
+    } else {
+      groupObjectCopy[groupField] = [item];
+    }
```
NikitaStambul, line 14: You can remove else statement like this:

```suggestion
    if (!Object.prototype.hasOwnProperty.call(groupObjectCopy, groupField)) {
      groupObjectCopy[groupField] = [];
    }
    
    groupObjectCopy[groupField].push(item);
```

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T) : GroupsMap<T> {
+  return items.reduce((map:GroupsMap<T>, item: T) => {
+    const groupField = item[key] as string;
+    const groupObjectCopy: GroupsMap<T> = { ...map };
```
NikitaStambul, line 8: I could be wrong but don't really sure that you need copy here

NikitaStambul (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T) : GroupsMap<T> {
+  return items.reduce((map:GroupsMap<T>, item: T) => {
```
sergii-nosachenko, line 6:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
  return items.reduce((map: GroupsMap<T>, item: T) => {
```
Enable linter.

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+    const groupField = item[key] as string;
```
sergii-nosachenko, line 7: Convert value to string instead of type casting.

src/groupByKey.ts
sergii-nosachenko, line 6: You can give a better name. I don't understand why it is called a `map`.

src/groupByKey.ts
sergii-nosachenko, line 14:
```suggestion
   
    if (!{}.hasOwnProperty.call(groupObjectCopy, groupField)) {
      // create property with empty array
    };
    
    groupObjectCopy[groupField].push(item);
```

sergii-nosachenko (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((groupsMap: GroupsMap<T>, item: T) => {
+    const groupField = String(item[key]);
+    const groupObjectCopy: GroupsMap<T> = { ...groupsMap };
+
+    if (!{}.hasOwnProperty.call(groupObjectCopy, groupField)) {
+      groupObjectCopy[groupField] = [];
+    }
```
sergii-nosachenko, line 12: You can also use the [Nullish coalescing assignment (??=)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) here:
```suggestion
    groupObjectCopy[groupField] ??= [];
```

PR #750, opened 2023-03-07
solution

NikitaStambul:


timoxa2t (approved):
Good work

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.reduce((acc, item) => {
```
sergii-nosachenko, line 8:
```suggestion
  return items.reduce((acc, item) => {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+    const groupKey = item[key] as string;
```
sergii-nosachenko, line 9: Convert value to a string instead of type casting.

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+
+    if (!(groupKey in acc)) {
+      acc[groupKey] = [];
+    }
```
sergii-nosachenko, line 13: Use [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) or [OR assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment).

sergii-nosachenko (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((acc: GroupsMap<T>, item) => {
+    const groupKey = String(item[key]);
+
+    acc[groupKey] ||= [];
```
OleksiiBlezniuk, line 9: Wow, that is next level 🔥

OleksiiBlezniuk (approved):

PR #752, opened 2023-03-08
function for grouping by key was created

rashyd-hasratov:


src/groupByKey.ts
```diff
@@ -1,7 +1,19 @@
 type GroupsMap<T> = {
-  [key: string]: T[];
+  [key: string | number]: T[];
```
sergey-lipikhin, line 2: Why do you need `number` type if you explicitly convert `itemKeyValue` to `string`

src/groupByKey.ts
```diff
@@ -1,7 +1,19 @@
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedByKey: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
```
sergey-lipikhin, line 8: You can do the same with `reduce` :)

sergey-lipikhin (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((grouped: GroupsMap<T>, item): GroupsMap<T> => {
+    const groupedCopy = { ...grouped };
+    const itemKeyValue = item[key] as string;
```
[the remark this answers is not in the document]
OleksiiBlezniuk: Better to avoid using `as` if it's possible to not using it

OleksiiBlezniuk (approved):

PR #754, opened 2023-03-08
GroupByKey task done

MarkBabich:


vladyslav-mosor (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+  const keysSet: Set<string> = new Set(items.map((item) => String(item[key])));
+  const keys: Array<string> = Array.from(keysSet);
+
+  keys.reduce((groupedItem, uniqKey): GroupsMap<T> => {
```
sergii-nosachenko, line 10:
```suggestion
  return items.reduce((groupedItem, item): GroupsMap<T> => {
```
You need to accumulate the resulting object in reduce. Just get groupBy value from every item (`String(item[key])`) and append it to the accumulator.

sergii-nosachenko (changes requested):

Andr1yK (approved):

PR #755, opened 2023-03-08
Solution

Ukrainiane-panda:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T & string): GroupsMap<T> {
```
Volodymyr-Melnyk-Fron10der, line 5: image
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+  const groups: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    const groupKey: T [keyof T & string] = item[key];
```
Volodymyr-Melnyk-Fron10der, line 9: `keyof T` - it's a type of certain key, but `item[key]` - it's not a key, it's a value
```suggestion
    const groupKey: string = item[key];
```

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+    if (groups[groupKey as string]) {
+      groups[groupKey as string].push(item);
+    } else {
+      groups[groupKey as string] = [item];
+    }
```
Volodymyr-Melnyk-Fron10der, line 15: 1. it's better to check an existing of the key in the object, not a value of the key
2. remove alias
```suggestion
    if (groupKey in groups) {
      groups[groupKey].push(item);
    } else {
      groups[groupKey] = [item];
    }
```

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there!

PR #758, opened 2023-03-09
GroupByKey

andriibezkorovainyi:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
+    const item: T = items[i];
+    const groupBy: keyof GroupsMap<T> = String(item[key]);
+
+    groupedItems[groupBy]
+      = items.filter((element) => String(element[key]) === groupBy);
```
OleksiiBlezniuk, line 13:
```suggestion
    groupedItems[groupBy] = items.filter(
      (element) => String(element[key]) === groupBy
    );
```

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
```
OleksiiBlezniuk, line 8: Would be nice to use array methods instead

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+    const item: T = items[i];
+    const groupBy: keyof GroupsMap<T> = String(item[key]);
```
OleksiiBlezniuk, line 10:
```suggestion
    const groupBy = String(item[key]);
```

OleksiiBlezniuk: Key of GroupsMap<T> is string, so no need to specify this type here since you converted this value to string

OleksiiBlezniuk (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const groupBy = String(item[key]);
+
+    groupedItems[groupBy] = items.filter(
+      (element) => String(element[key]) === groupBy,
+    );
```
sergii-nosachenko, line 13: Not an optimal way. You will loop multiple times over a whole array in case of duplicate keys. Better add a check if `groupBy` prop is already present in the object to avoid an unnecessary array loop.

sergii-nosachenko (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const groupBy = String(item[key]);
+
+    if (groupBy in groupedItems) {
```
sergii-nosachenko, line 11:
```suggestion
    if ({}.hasOwnProperty.call(groupedItems, groupBy) {
```
The proper way to check this

sergii-nosachenko (changes requested):

OleksiiBlezniuk (approved):

PR #763, opened 2023-03-09
Solution

VadymTsyndra:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+  items.forEach((item) => {
+    if (!obj[item[key]]) {
+      obj[item[key]] = [item];
+    } else {
+      obj[item[key]].push(item);
+    }
+  });
+
+  return obj;
 }
```
SerhiiKirik, line 17: you should use generic here
```suggestion
export function groupByKey<T>(items: any[], key: string): GroupsMap<T> {
  const obj: GroupsMap<T> = {};

  items.forEach((item) => {
    if (!obj[item[key]]) {
      obj[item[key]] = [item];
    } else {
      obj[item[key]].push(item);
    }
  });

  return obj;
}
```

SerhiiKirik (changes requested):
Well done 🔥 
Let's fix some problem 🧑‍💻

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: T): GroupsMap<T> {
+  const obj: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    if (!obj[item[key]]) {
+      obj[item[key]] = [item];
+    } else {
+      obj[item[key]].push(item);
+    }
```
sTorba24, line 13:
```suggestion
    if (!obj[item[key]]) {
      obj[item[key]] = [item];
      return;
    } 
    
    obj[item[key]].push(item);
```

sTorba24 (approved):
GJ :)

PR #766, opened 2023-03-11
solution for ts_group_by_key

petrovskyi-yevhen:


src/groupByKey.ts
```diff
@@ -2,6 +2,10 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
```
TarasHoliuk, line 5: `string` type is not enough specified: use [keyof](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html)

TarasHoliuk (changes requested):

PR #781, opened 2023-03-23
GroupByKey

Oleksandr-Maslo:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
```
OleziO, line 8: Replace `i += 1` with `i++`

OleziO (approved):

src/groupByKey.ts
Oleksandr-Maslo: then the linter does not pass

error Unary operator '++' used

PR #785, opened 2023-03-24
Solution

IrynaK-a:


src/groupByKey.ts
```diff
@@ -2,6 +2,23 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(
+  items: Array<object>,
```
agniya-i, line 6:
```suggestion
  items: Array<T>,
```

better use `T`, `object` can be a little bit confusing

src/groupByKey.ts
```diff
@@ -2,6 +2,23 @@ type GroupsMap<T> = {
...
+  key: string,
```
agniya-i, line 7:
```suggestion
  key: keyof T,
```

it would be better to convert key to string when you iterate through the loop

src/groupByKey.ts
```diff
@@ -2,6 +2,23 @@ type GroupsMap<T> = {
...
+): GroupsMap<object> {
```
agniya-i, line 8:
```suggestion
): GroupsMap<T> {
```

agniya-i (changes requested):

OleziO (approved):

PR #796, opened 2023-04-02
added function groupByKey

Esteban4769:


PavloReutskiy (approved):
👍

SerhiiKirik (changes requested):
Well done 🚀 
Let's try to improve your solution 🧑‍💻

igorkruz (approved):
👍

Olexandr6 (approved):
Well done🔥!

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T extends object>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  return items.reduce((acc, item) => {
+    const keyValue = item[key] as keyof GroupsMap<T>;
```
Olexandr6, line 10: Avoid using `as` in TS
```suggestion
    const keyValue = String(item[key]);
```

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+
+    if (!acc[keyValue]) {
+      acc[keyValue] = [];
+    }
+
+    acc[keyValue].push(item);
+
+    return acc;
+  }, {} as GroupsMap<T>);
```
Olexandr6, line 19: The same here, let's add type to `acc` to avoid using `as`

Olexandr6 (changes requested):
Sorry for the erroneous approval.
Now, let's work on improvements:

Kavon-Skela (approved):
Can't see any mistakes now. Good job!

PR #798, opened 2023-04-03
Group by Key TS v.0.1 - Solution

NedobezhkinOleksandr:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const objectsInGroup: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyObject = String(item[key]);
```
Volodymyr-Melnyk-Fron10der, line 9: Consider using a more descriptive variable name instead of `keyObject`. 🤔

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+    if (keyObject in objectsInGroup) {
```
Volodymyr-Melnyk-Fron10der, line 11: Consider using `hasOwnProperty` instead of `in` for better performance and to avoid issues with inherited properties. 🚀

src/groupByKey.ts
Volodymyr-Melnyk-Fron10der:
```suggestion
    const newKey = String(item[key]);
```

Volodymyr-Melnyk-Fron10der (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const objectsInGroup: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const newKey = String(item[key]);
+    const objProtoOfNewKey
+    = Object.prototype.hasOwnProperty.call(objectsInGroup, newKey);
```
anstsot, line 11: can be simplified
```suggestion
    const objProtoOfNewKey = objectsInGroup.hasOwnProperty(newKey);
```

anstsot (approved):
Good work, approved

PR #801, opened 2023-04-04
Solution

igorkruz:


AnyaLyaLya (approved):
GJ!

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T):GroupsMap<T> {
+  return items.reduce((acc: GroupsMap<T>, el) => {
```
Olexandr6, line 6:
```suggestion
  return items.reduce((acc: GroupsMap<T>, element) => {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+    const keyValue = el[key] as string;
```
Olexandr6, line 7: Avoid using `as` in TS
```suggestion
    const keyValue = String(el[key]);
```

Olexandr6 (changes requested):
Almost done👍 Now, let's work on improvements:

Olexandr6 (approved):

PR #804, opened 2023-04-05
Solution

AnyaLyaLya:


src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T extends Record<string, any>>(
```
BondarenkoDmitriy, line 5: i think instead of <T extends Record<string, any>> would be enough <T

BondarenkoDmitriy (approved):
GJ

src/groupByKey.ts
liliia-livandovskaya, line 5:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T>
```
do not use `any`, you need to specify the type

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
```
liliia-livandovskaya, line 8:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
+  return items.reduce((groups, item) => {
+    const value = item[key];
```
liliia-livandovskaya, line 10:
```suggestion
    const value = String(item[key]);
```

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((groups, item) => {
```
liliia-livandovskaya, line 6:
```suggestion
  return items.reduce((groups: GroupsMap<T>, item) => {
```

liliia-livandovskaya (approved):

PR #809, opened 2023-04-06
solution

Olekk17:


VladislavaT (approved):
Great job on writing this code! It's clear that you have a solid understanding of TypeScript's advanced type system and how to apply it to solve real-world problems.

The code defines a generic GroupsMap type that represents a mapping between string keys and arrays of values of type T. The groupByKey function takes an array of T items and a key that represents a property of the T type. It then uses the reduce function to iterate over each item in the array and group them based on the value of the specified key.

The code is very concise and easy to read, and it makes great use of TypeScript's type system to ensure type safety throughout the code. I also appreciate the use of descriptive variable names and comments to help make the code more understandable.

Overall, this is a great example of how to write clean, efficient, and type-safe TypeScript code. Well done!

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  return items.reduce((acc, item) => {
```
liliia-livandovskaya, line 9:
```suggestion
  return items.reduce((acc: GroupsMap<T>, item) => {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+    const value = <keyof GroupsMap<T>>item[key];
```
liliia-livandovskaya, line 10:
```suggestion
    const value = String(item[key]);
```

liliia-livandovskaya (changes requested):
GJ! Pay attention to the comments

SerhiiKirik (approved):
Great 🔥

PR #815, opened 2023-04-06
add solution

Pozdnya:


Hamuud (approved):
With such a code and such programming knowledge that you have, I think your first offer will be from Google😮‍💨, and once you finish React, you will generally write your own Google

liliia-livandovskaya (approved):

PR #821, opened 2023-04-06
add function groupByKey

Vlad-Lischenovskiy:


Stambuliak (approved):
agree

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedMap: GroupsMap<T> = {};
+
+  items.forEach((item) => {
```
liliia-livandovskaya, line 8: it's better to use `reduce` here

liliia-livandovskaya (approved):

PR #826, opened 2023-04-06
solution

Sbavdik:


src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((groups, item) => {
+    const value = item[key];
+    const group = groups[value as string] || [];
```
ymys1, line 8: https://mateacademy.slack.com/archives/C04N8TU8PN2/p1680773288486249

ymys1: Your code looks good! But you should correct this little thing) check link in previous message)

ymys1: Also there are some other 'as' in your code, you should correct it too

ymys1 (changes requested):
Good job, but need to be corrected little bit) Check comments

src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((groups, item) => {
+    const value = item[key];
+    const group = groups[String(value)] || [];
```
SerhiiKirik, line 8:
```suggestion
    const value = item[key];
    const key = String(value);
    const group = groups[key] || [];
```

src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
...
+
+    return {
+      ...groups,
+      [String(value)]: [...group, item],
+    };
+  }, {} as GroupsMap<T>);
```
SerhiiKirik, line 14: try to avoid as
```suggestion
  }, {});
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((groups, item) => {
+    const value = item[key];
+    const stringValue = String(value);
+    const group = groups[stringValue] || [];
+
+    return {
+      ...groups,
+      [String(value)]: [...group, item],
```
SerhiiKirik, line 13:
```suggestion
    const key = String(value);
    const group = groups[stringValue] || [];

    return {
      ...groups,
      [key]: [...group, item],
```

src/groupByKey.ts
SerhiiKirik: Can you fix it?

SerhiiKirik (changes requested):
Well done 🚀 
Pay attention, make a fix on the whole code 🧑‍💻

SerhiiKirik (changes requested):
Almost done 🚀 
Still not fixed 🧑‍💻 
image

Sbavdik:
> Almost done 🚀 Still not fixed 🧑‍💻 image

My bad, just didn't see that. Too much comments🤯

Olexandr6 (approved):
Well done🔥!

PR #840, opened 2023-04-07
Solution

Zibi95:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items:T[], key: keyof T):GroupsMap<T> {
+  const seed:GroupsMap<T> = {};
```
Radoslaw-Czerniawski, line 6: No need for creating a separate variable for this, you could have on line 8
```
items.reduce((accumulator: GroupsMap<T>, item) => {
```
and then on line 18 just pass an empty object `{}`
or you could have instead on line 8 casted the object on line 18 like so
```
 }, {} as GroupsMap<T>); -> i would say this is less preferable
```

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+  return items.reduce((accumulator, item) => {
+    const groupName:string = String(item[key]);
```
Radoslaw-Czerniawski, line 9: Generally in typescript we should rely as much as possible on type inference (typescript figuring out on its own the types) rather than explicit types, especially here where it obviously is a string. There may be an exception to this as some reviewers like to have explicit types, especially on return types of functions (I would say typing return types of functions is a thing that team of developers agree or not in their coding standards)

Radoslaw-Czerniawski (approved):
Good job Sylwester, if you will be bored give me a message I will send you an additional task.

PR #848, opened 2023-04-10
solution

toljan1:


IvanFesenko (approved):
Approved. But for practice you can try do it with `Array.reduce()`

PR #856, opened 2023-04-20
Group by key solution

mykola-chernysh:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((acc: GroupsMap<T>, elem) => {
```
polosanya, line 6: Consider adding a type annotation for the `elem` parameter in the reduce callback function for better type safety. Example: `elem: T`

polosanya (approved):
Good job!

PR #864, opened 2023-04-27
Solution

SuaBy514:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
+export function groupByKey<
+  T, R extends keyof T
+>(
+  items: T[], key: R,
+): GroupsMap<T> {
   // write code here;
+  const result: GroupsMap<T> = {};
```
d-wasilewski, line 11:
```suggestion
  const groupedByKey: GroupsMap<T> = {};
```

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+
+  for (let i: number = 0; i < items.length; i += 1) {
```
d-wasilewski, line 13: I don't think you need to type this. Some of the types are automatically infered
```suggestion
  for (let i = 0; i < items.length; i += 1) {
```

d-wasilewski, line 13: maybe use forEach method?

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
+export function groupByKey<
+  T, R extends keyof T
+>(
+  items: T[], key: R,
```
d-wasilewski, line 8: You need only 1 generic type here
```suggestion
export function groupByKey<T>(
  items: T[], key: keyof T,
```

d-wasilewski (changes requested):

src/groupByKey.ts
SuaBy514: Easier with for, but I will try

d-wasilewski (approved):

PR #877, opened 2023-05-16
add groupByKeys solution

DidenkoDenis306:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyOfItem: string = String(item[key]);
```
dinaminahi, line 9: Consider using 'as string' instead of 'String()' for type casting to string.

dinaminahi, line 9: You don`t have to specify string type here
https://www.typescripttutorial.net/typescript-tutorial/typescript-type-inference/

dinaminahi (approved):
Good job 👍 
But check a comment

PR #879, opened 2023-05-17
function groupByKey was added

Olena-Shevchenko:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
+  let arr: string[] = [];
+  const result: GroupsMap<T> = {};
+
+  arr = items.map((item) => item[key]);
+  arr = [...new Set(arr)];
+
+  arr.forEach((item) => {
+    result[item] = items.filter((one) => one[key] === item);
```
Anastasiia-Svintsova, line 13: Use a more descriptive variable name instead of `one`.

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
+  let arr: string[] = [];
+  const result: GroupsMap<T> = {};
+
+  arr = items.map((item) => item[key]);
```
Anastasiia-Svintsova, line 9: Instead of using a separate array `arr`, you can directly create the unique keys in the resulting object.

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
+  let arr: string[] = [];
```
Anastasiia-Svintsova, line 6: You don't need to create a new array here

Anastasiia-Svintsova (changes requested):

witflash (approved):
Well done 👍

PR #894, opened 2023-05-31
solution

vpdrabynko:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  for (let i: number = 0; i < items.length; i += 1) {
```
polosanya, line 8: Consider using a 'for of' loop or 'forEach' method instead of a traditional 'for' loop for better readability.

polosanya (approved):
👍

PR #895, opened 2023-05-31
add task solution

KatyaZaremska:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groups: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const groupKey: string = String(item[key]);
```
33Orange, line 9: TS will infer it from usage
```suggestion
    const groupKey = String(item[key]);
```

33Orange (approved):
Well done !

PR #897, opened 2023-06-01
task solution

hekaq:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groups: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const value: string = String(item[key]);
```
naziamoff, line 9: should work without explicitly mentioning `string` data type - TS will infer it from usage
```suggestion
    const value = String(item[key]);
```

naziamoff (approved):
🔥 
See one suggestion

PR #898, opened 2023-06-01
Add task solution

IrynaBalandiukh:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const sortKey: string = String(item[key]);
```
naziamoff, line 9: not necessary to explicitly mention types like string - typescript will infer it from usage
```suggestion
    const sortKey = String(item[key]);
```

naziamoff (approved):
good job 🔥

PR #899, opened 2023-06-01
Solution

julshvets:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item: T): void => {
+    const currentKeyValue:string = String(item[key]);
+
+    if (currentKeyValue in result) {
+      result[currentKeyValue].push(item);
+    } else {
+      result[currentKeyValue] = [item];
+    }
+  });
```
33Orange, line 19: Should work without unnecessary type descriptions. Since the iteration is already over an array of items with types T - TypeScript understands that each item will be T. There is also no need to describe the action of the forEach method, since this is a built-in method and TS knows what it does

```suggestion
  items.forEach((item) => {
    const currentKeyValue= String(item[key]);

    if (currentKeyValue in result) {
      result[currentKeyValue].push(item);
    } else {
      result[currentKeyValue] = [item];
    }
  });
```

33Orange (approved):
Good job.


Pay attention to the comment

src/groupByKey.ts
julshvets: Thank you!
Done.

PR #900, opened 2023-06-01
solution

Yuliya1907:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const itemKey = item[key] as string;
```
naziamoff, line 9: is this type conversion necessary? You should avoid type conversion where possible. Let's just convert item[key] to the string value instead of converting the type

naziamoff (approved):
good job, see one suggestion

PR #903, opened 2023-06-01
add groupByKey task solution

opiquem:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupMap: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const valueSort: string = item[key].toString();
```
naziamoff, line 9: not a mistake, but it is usually converted this way
```suggestion
    const valueSort: string = String(item[key]);
```

naziamoff, line 9: also - not necessary to explicitly add `string` type, as Typescript will infer it from usage

naziamoff (approved):
good job 🔥 
See some suggestions

PR #905, opened 2023-06-01
Solution

annholovko:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouppedItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const itemKey = item[key] as string;
```
naziamoff, line 9: let's not convert types unless we really need to. You can convert the value instead

naziamoff (changes requested):
not bad, but it is important not to convert types where it is possible

33Orange (approved):
Well done

PR #906, opened 2023-06-01
add solution

ulositska:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items:T[], key: keyof T): GroupsMap<T> {
+  const result:GroupsMap<T> = {};
```
naziamoff, line 6: better not to use variable names like `result`. `groupedItems` would work better in this case

naziamoff (approved):
Good job! See one suggestion

PR #908, opened 2023-06-01
ts_group_by_key

YelyzavetaMolochynska:


src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
...
+export function groupByKey<T>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  return items.reduce((grouped: GroupsMap<T>, item: T) => {
+    const value = String(item[key]);
+
+    return {
+      ...grouped, [value]: [...(grouped[value] || []), item],
+    };
```
polosanya, line 14:
```suggestion
    return {
      ...grouped, 
      [value]: [...(grouped[value] || []), item],
    };
```
It would be also great to create a variable for an array to make your code a bit cleaner

polosanya (approved):
👍

PR #909, opened 2023-06-01
Solution

bohdan-mykhailenko:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((groups: GroupsMap<T>, item: T) => {
+    const itemKey = item[key] as string;
```
naziamoff, line 7: let's not convert the type - let's convert the value to type instead

naziamoff (approved):
nice idea to use reduce here 🔥

PR #911, opened 2023-06-01
Solution

amashutikov:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const sortKey: string = String(item[key]);
```
naziamoff, line 9: should work without explicit mention of the string type, TS will infer it from usage
```suggestion
    const sortKey = String(item[key]);
```

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
```
naziamoff, line 6:
```suggestion
  const groupedItems: GroupsMap<T> = {};
```

naziamoff (approved):

PR #913, opened 2023-06-01
add solution

ivankhrsh:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupsMap: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const sortKey : string = String(item[key]);
```
naziamoff, line 9: should work just fined without explicit mention of the type. TS will infer it from usage
```suggestion
    const sortKey = String(item[key]);
```

naziamoff (approved):
gj 👍

PR #916, opened 2023-06-02
Solution

PanyaPrimal:


src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): Record<string, T[]> {
+  return items.reduce((grouped: GroupsMap<T>, item: T) => {
+    const itemKey = item[key] as unknown as string;
```
33Orange, line 7: Change key and key type instead of using `as`

```suggestion
    const itemKey = String(item[key])
```

33Orange (changes requested):
Almost done

src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): Record<string, T[]> {
+  return items.reduce((grouped: GroupsMap<T>, item: T) => {
+    const itemKey = String(item[key]);
+
+    return {
+      ...grouped,
+      [itemKey]: grouped[itemKey] ? [...grouped[itemKey], item] : [item],
```
silchyn, line 11:
```suggestion
      [itemKey]: grouped[itemKey]
        ? [...grouped[itemKey], item]
        : [item],
```

silchyn (approved):

PR #917, opened 2023-06-02
Solution

Mykola-Maik:
Hi! All tests were completed, but I have one Type error  "Type 'T[keyof T]' cannot be used to index type 'GroupsMap<T>" 
and I didn't get how to fix it, or why this error came up? If it's ok that you answer here in comment then good, but if I need to go to Q&A session, just say)) 
Thank you!

### DIFF
--- a/src/groupByKey.ts
+++ b/src/groupByKey.ts
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result = {} as GroupsMap<T>;
+
+  items.forEach((item) => {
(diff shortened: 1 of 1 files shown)


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result = {} as GroupsMap<T>;
+
+  items.forEach((item) => {
+    const group = item[key];
```
33Orange, line 9:
```suggestion
    const group = String(item[key]);
```

33Orange (changes requested):
Almost done.

This will fix your error

silchyn (approved):

PR #920, opened 2023-06-02
ts_group_by_key

vintwp:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  // const group = {} as GroupsMap<T>;
```
VitaliyBondarenko1982, line 6: remove commented code

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+
+  const group = items.reduce((books: GroupsMap<T>, book) => {
+    const value = String(book[key]);
+
+    if (!books[value]) {
+      // eslint-disable-next-line no-param-reassign
```
VitaliyBondarenko1982, line 12: for avoid this elsint rule
you can assign books to some variable and work with this variable
or use ternay operator and destructuring
```
 return books[value] ? {
      ...books, [value]: [...books[value], book],
    } : {
      ...books, [value]: [book],
    };

File: src/groupByKey.ts [hunk 3]
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  // const group = {} as GroupsMap<T>;
+
+  const group = items.reduce((books: GroupsMap<T>, book) => {
```
VitaliyBondarenko1982, line 8:
```suggestion
  return items.reduce((books: GroupsMap<T>, book) => {
```
you can immediately return result of reduce

VitaliyBondarenko1982 (changes requested):

OleziO (approved):

PR #925, opened 2023-06-05
Solution

Ir-ra:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groped: GroupsMap<T> = {};
```
lerastarynets, line 6: I suppose, you wanted to use `grouped` here)

lerastarynets (approved):
Good job, I like that your solution is that short :)

PR #931, opened 2023-06-12
Solution

kateryna-stemblivska:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedObject: GroupsMap<T> = {};
+
+  items.forEach((el) => {
```
lerastarynets, line 8: Please consider using more descriptive naming instead of `el`)

lerastarynets (approved):
Good job) Just one small think that should be improved

PR #933, opened 2023-06-14
solution

Gosyav:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const value = item[key];
+
+    if (!result[<string>value]) {
+      result[<string>value] = [];
+    }
```
VitaliyBondarenko1982, line 13:
```suggestion
    const value = String(item[key]);

    if (!result[value]) {
      result[value] = [];
    }
```
better to make string one time

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
```
VitaliyBondarenko1982, line 8:
```suggestion
  return items.reduce<GroupsMap<T>>((item) => {
```
consider to use reduce method and return its result

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #942, opened 2023-06-18
Task resolved

read-my-code:


src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: [], key: string): GroupsMap<object> {
+  let obj: GroupsMap<object> = {};
```
SerhiiKirik, line 6: why do you need `let` here? You can just mutate your obj

TarasHoliuk (approved):

PR #943, opened 2023-06-18
ts_group_by_key

ArtemKravcov85:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey <T>(item: T[], key: keyof T): GroupsMap<T> {
```
polosanya, line 5:
```suggestion
export function groupByKey <T>(items: T[], key: keyof T): GroupsMap<T> {
```
Actually, it's an array of items, name it correctly

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+  const result: GroupsMap<T> = {};
```
polosanya, line 6: It is better to avoid naming variables `result`. Try to specify what exactly the result is

polosanya (changes requested):
Almost perfect!

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey <T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result = {} as GroupsMap<T>;
```
polosanya, line 6:
```suggestion
  const groupedItems = {} as GroupsMap<T>;
```

polosanya (approved):
👍

ArtemKravcov85:
Thanks!

пн, 19 июн. 2023 г., 11:57 Oleksandr Polotniuk ***@***.***>:

PR #944, opened 2023-06-18
solution

vlkzmn:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
```
polosanya, line 6: It is better not to use `result` as a name. It would be great to specify what exactly the result is

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item: any) => {
```
polosanya, line 8: Avoid using `any` in TypeScript, it's a bad practice. Consider setting the correct type or just remove the typization here, as `items` is an array of T and it's clear what element of the array is

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+    if (!result[item[key]]) {
+      result[item[key]] = [item];
+    } else {
+      result[item[key]].push(item);
```
polosanya, line 12: Create a variable for `item[key]` as it duplicates three times

polosanya (changes requested):
Almost perfect!

polosanya (approved):
👍

PR #946, opened 2023-06-20
Solution

Viktoriia6666:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((obj: GroupsMap<T>, item: T) => {
+    const acc = obj;
```
loralevitska, line 7: Don't cut variable names. What does it mean `acc`? 
Try to use more meaningful variable naming

loralevitska (changes requested):
Almost perfect) Let's improve

loralevitska (approved):
Well done)

PR #950, opened 2023-06-21
Group by key

ilona-romaniuk:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedObj: GroupsMap<T> = {};
+
+  items.forEach((obj: T) => {
+    const property: string = `${obj[key]}`;
```
lerastarynets, line 9: Pls consider using a more descriptive variable name instead of 'obj'. For example, 'item' or 'element' would be more suitable.

lerastarynets, line 9: It's unnecessary to use template literals for converting obj[key] to a string. Use 'String(obj[key])' instead.

lerastarynets (changes requested):
Nice start! Left few comments on how to improve your code

lerastarynets (approved):
All requirements are met.  Keep up the great work!

PR #954, opened 2023-06-24
solution

Absolution26:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce<GroupsMap<T>>((acc, item) => {
+    const keyValue = String(item[key]);
```
lerastarynets, line 7: Consider using a more descriptive variable name instead of 'acc' for better readability. For example, you can use 'groups' or 'groupedItems'.

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+
+    if (!acc[keyValue]) {
+      acc[keyValue] = [];
```
lerastarynets, line 10: You can use optional chaining and nullish coalescing to simplify this line. Replace this block with: `acc[keyValue] = acc[keyValue] ?? [];`

lerastarynets (changes requested):
Nice start! Left few comments on how to improve your code

src/groupByKey.ts
Absolution26: Sry, cannot change variable name, error "Assignment to property of function parameter 'groups',  no-param-reassign". Also check 'gro' and 'groupedItems'
But work with 'acc', I don't know why

OleziO (approved):

PR #956, opened 2023-06-28
add task solution

MelnychukVol:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const newObject: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyForObj = item[key] as string;
```
volodymyr-soltys97, line 9: Type assertion to 'string' is potentially unsafe. If the value of 'key' is not a string, it can lead to runtime errors. Consider improving type safety by using a type guard or a different approach to handle potential non-string values.

volodymyr-soltys97 (approved):
Good work 👍

PR #957, opened 2023-06-28
add task solution

ilona-borkovska:


src/groupByKey.ts
```diff
@@ -1,7 +1,40 @@
...
+
+interface Book {
+  author: string;
+  country: string;
+  imageLink: string;
+  language: string;
+  pages: number;
+  title: string;
+  year: number;
+}
```
anstsot, line 21: try not to use specific interfaces

src/groupByKey.ts
```diff
@@ -1,7 +1,40 @@
...
+
+export function groupByKey(
+  items: Array<Student | Book>,
```
anstsot, line 24: change to the common type to add possibility array of any objects

anstsot (changes requested):
Let’s add a few fixes to improve your code.

meforoma (changes requested):
{technical reject}

And678 (approved):
Great, approve :heavy_check_mark:

PR #964, opened 2023-06-30
Solution

Nataliia95254:


src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: [], key: string): GroupsMap<{}> {
+  const result: GroupsMap<{}> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
+    const sort = items.filter((obj) => obj[key] === items[i][key]);
```
VitaliyBodnar, line 9: 'sort' is a reserved word like a map, var, or function 
so, make sense to rename it

VitaliyBodnar (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: [], key: string): GroupsMap<{}> {
+  const result: GroupsMap<{}> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
```
OleziO, line 8: If this does not cause a linter error - replace `i += 1` with `i++` (If there is an error, write it under my comment)

OleziO (changes requested):

src/groupByKey.ts
Nataliia95254: i++ linter error (Unary operator '++' used.)

OleziO (approved):

PR #965, opened 2023-06-30
add task solution

denys2:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.forEach((element) => {
+    const value = String(element[key]);
```
lerastarynets, line 9: The use of String(element[key]) is not necessary. Consider removing the String() wrapper.

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+    if (!grouped[value]) {
```
lerastarynets, line 11: Just fyi
`!grouped[value]` checks the truthiness of the value stored at the `value` property in the `grouped` object. It returns true if the value is falsy (e.g., undefined, null, false, 0, empty string, etc.), and false if it is truthy.
While `!(value in grouped)` uses the in operator to check if the `value` exists as a property in the `grouped` object. It returns true if the property does not exist and false if it exists.

In terms of performance, `!(value in grouped)` may be slightly more efficient because it directly checks for the existence of the property using the in operator. `!grouped[value]` relies on the truthiness of the value, which involves an additional property access.

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #966, opened 2023-07-01
solution

DanyloPasko:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((obj: GroupsMap<T>, item: T) => {
+    const group = obj;
```
volodymyr-soltys97, line 7: You shouldn't assign the `obj` to `group` directly as it just copies the reference, not the object itself. Any changes to `group` will reflect in `obj`. Use spread operator to clone the object.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

volodymyr-soltys97 (approved):
Nice work 👍

PR #968, opened 2023-07-01
solution

vlad-fil:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item) => {
+    const value = item[key] as string;
+
+    if (!result[value]) {
+      result[value] = [];
+    }
+
+    result[value].push(item);
+  });
```
etojeDenys, line 16: you can rewrite it using `reduce`, it will improve readability

etojeDenys (approved):
lgtm

PR #970, opened 2023-07-02
task done

tecHamster:


src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((acc: GroupsMap<T>, item) => {
+    const keyValue = String(item[key]);
```
lerastarynets, line 7: Unnecessary conversion of `item[key]` to string. The key of an object can be not only a string, but also a symbol. Moreover, JavaScript automatically converts number keys to strings. Consider removing the `String()` wrapper.

lerastarynets (approved):
You're doing a fantastic job, keep going! Left one tiny suggestion for the improvement

PR #972, opened 2023-07-03
Solution

vitalii-fedusov:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+  const obj: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    const objKey = item[key] as string;
+
+    if (obj[objKey]) {
+      obj[objKey].push(item);
+    } else {
+      obj[objKey] = [item];
+    }
```
etojeDenys, line 15: looks good, but we can simplify it a little bit

```suggestion
    if (!obj[objKey]) {
       obj[objKey] = [];
    } 
       
    obj[objKey].push(item);
```

etojeDenys (changes requested):

Viktor-Kost (approved):

PR #981, opened 2023-07-07
group books by keys

ValeraViachalo:


IvanFesenko (approved):
Interesting solution.
For practice try resolve task with `Array.reduce`

PR #983, opened 2023-07-07
Solution

denlubn:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const gropedItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const value: string = String(item[key]);
```
volodymyr-soltys97, line 9: You don't need to convert the key value to a string explicitly as it will be automatically converted when used as an object key.

volodymyr-soltys97 (approved):
Well done 🔥

PR #984, opened 2023-07-08
done

diana-tuz:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+export function groupByKey(items: [], key: string): GroupsMap<Object> {
+  const group: GroupsMap<Object> = {};
+
+  items.forEach((item) => {
+    if (item[key]) {
+      if (!group[item[key]]) {
+        group[item[key]] = [];
+      }
+    }
+    group[item[key]].push(item);
```
volodymyr-soltys97, line 14: Pushing the item into the group[item[key]] array is done outside of the if block that checks if 'item[key]' exists. This could potentially lead to errors if 'item[key]' doesn't exist. You should move this line inside the if block.

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: [], key: string): GroupsMap<Object> {
+  const group: GroupsMap<Object> = {};
+
+  items.forEach((item) => {
+    if (item[key]) {
```
volodymyr-soltys97, line 9: The check if 'item[key]' exists could potentially be problematic if the value is 0 or any other falsy value. Try to be more specific with your checks.

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+      if (!group[item[key]]) {
+        group[item[key]] = [];
```
volodymyr-soltys97, line 11: Also you can create a variable for `item[key]` and use it

volodymyr-soltys97 (changes requested):
Good job
But let's make your code perfect

nazarmatsevych (approved):

PR #985, opened 2023-07-08
add task solution

Vayts:


alexander-ignatow (changes requested):
Very good - just a minor typescript polish needed.

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+  const resultObject: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const objKey = item[key] as string;
+
+    if (!resultObject[objKey]) {
+      resultObject[objKey] = [item];
+    } else {
+      resultObject[objKey].push(item);
+    }
```
BudnikOleksii, line 15: Prefer don't use else, you can create an empty array if doesn't exist and in both cases just push an item.

BudnikOleksii (approved):
Good job!

PR #988, opened 2023-07-09
Solution

AntonKabanets:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const group: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const itemKey = String(item[key]);
```
VitaliyBodnar, line 9: You don't need to convert the key value to a string explicitly as it will be automatically converted when used as an object key.

VitaliyBodnar (changes requested):
Well done )

PR #1006, opened 2023-07-13
add solution

vetal-hovenko:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: Object[], key: string): GroupsMap<T> {
+  const group: GroupsMap<T> = {};
+
+  items.forEach((object: any) => {
```
polosanya, line 8: Avoid using `any`. Specify what exactly the object is

polosanya (changes requested):
Almost perfect!

src/groupByKey.ts
```diff
@@ -2,6 +2,25 @@ type GroupsMap<T> = {
...
+  [key: string]: string | number;
+};
+
+export function groupByKey<T>(
+  items: ObjectForSorting[],
+  key: string,
+): GroupsMap<T> {
+  const group: GroupsMap<T> = {};
+
+  items.forEach((object: ObjectForSorting) => {
```
volodymyr-soltys97, line 15: The 'forEach' loop can be replaced with 'reduce' for better performance and readability. The 'reduce' function is more suitable for this case because it can transform an array into an object. Additionally, it will ensure compliance with the 'array-callback-return' rule in ESLint as 'reduce' requires a return value.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

src/groupByKey.ts
```diff
@@ -2,6 +2,24 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+type ObjectForSorting = {
```
anastasiiavorobiova, line 5:
```suggestion
type ObjectForSorting = Record<string, string | number>
```
There are records in the typescript for such purpose

src/groupByKey.ts
```diff
@@ -2,6 +2,24 @@ type GroupsMap<T> = {
...
+  [key: string]: string | number;
+};
+
+export function groupByKey<T>(
+  items: ObjectForSorting[],
+  key: string,
+): GroupsMap<T> {
+  return items.reduce((acc: GroupsMap<T>, object: ObjectForSorting) => {
```
anastasiiavorobiova, line 13: You can use the Record here instead of defining the type

anastasiiavorobiova (changes requested):
Great solution! Added comments to fix

nazarmatsevych (approved):

PR #1021, opened 2023-07-24
Solution

reseneweb:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function
```
lerastarynets, line 5: Function keyword and function name should be on the same line. It is better to change it to 'export function groupByKey'

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+groupByKey<T, K extends keyof T>(items: T[], key: K): GroupsMap<T> {
+  const result: GroupsMap<T> = items.reduce((acc, ch) => {
+    const keyValue = ch[key] as unknown as string;
```
lerastarynets, line 8: Avoid using 'as unknown as string'. It's better to use type guards or type assertions properly to avoid any possible type errors.

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+    if (acc[keyValue] === undefined) {
```
lerastarynets, line 10: pls [see falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
```suggestion
    if (!acc[keyValue]) {
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey
+<T, K extends keyof T>(items: T[], key: K): GroupsMap<T> {
```
lerastarynets, line 6: you can simplify this
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
```

lerastarynets (approved):
Keep up the great work!

PR #1025, opened 2023-07-25
solution

Ilya-Osoben:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
   // write code here;
+
+  return items.reduce((newGroup, item) => {
+    const currentKey = item[key] as string;
```
TarasHoliuk, line 9: don't use typecasting here, transform it into a string explicitly

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+    const group = newGroup[currentKey] || [];
+
+    return {
+      ...newGroup,
```
TarasHoliuk, line 13: this variable name doesn't describe its value

TarasHoliuk (changes requested):

PR #1026, opened 2023-07-27
Implement task groupByKey.ts

oleksandrshtonda:


src/groupByKey.ts
```diff
@@ -2,6 +2,29 @@ type GroupsMap<T> = {
...
+
+  type ValueFromObject = ObjectToSort[Key];
+
+  const valuesOfKey: Array<ValueFromObject> = items.map(
+    (item: ObjectToSort) => item[key],
+  );
+  const uniqueValuesOfKey: Set<ValueFromObject> = new Set([...valuesOfKey]);
+  const sortedObjects: GroupsMap<ObjectToSort> = {};
+
+  uniqueValuesOfKey.forEach((uniqueValueOfKey: ValueFromObject) => {
```
volodymyr-soltys97, line 21: The filter operation inside the forEach loop results in a quadratic time complexity. This can be avoided by using a single reduce function to create the grouped object.

src/groupByKey.ts
```diff
@@ -2,6 +2,29 @@ type GroupsMap<T> = {
...
+  items: ObjectToSort[],
+  key: Key,
+): GroupsMap<ObjectToSort> {
+  if (items.length === 0) {
+    return {};
+  }
+
+  type ValueFromObject = ObjectToSort[Key];
+
+  const valuesOfKey: Array<ValueFromObject> = items.map(
```
volodymyr-soltys97, line 15: The map operation in line 16 is unnecessary. You're creating an array of values only to create a set later. This is inefficient as it results in two iterations over the items array.

volodymyr-soltys97 (changes requested):
The solution looks big and a bit confusing, try to simplify it to make it easier for other developers to read.
You can create an empty object at the beginning and then use only a forЕach loop to iterate, and you will write what you need into the object

PR #1035, opened 2023-08-07
Solution

KarolinaWiktoria:


src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((result, item) => {
+    const keyValue = String(item[key]);
+
+    return {
+      ...result,
+      [keyValue]: [...(result[keyValue] || []), item],
```
darokrk, line 11: The spread operator is used inside a loop, which can lead to performance issues due to creating a new object on each iteration. Consider using a different approach to add properties to the 'result' object.

We can just simplify this logic, please remove your `return {}` and try this one to be:

```
if (!result[keyValue]) {
   result[keyValue] = [];
}
result[keyValue].push(item);

return result;
```

darokrk (changes requested):
Not bad, please look at my comment to make code improvements 😃

darokrk (approved):
Well done 👏🏽 , the solution with `reduce()` function was fine, but `for loop` is also fine 👍🏽  good job 🔥 😃 !

KarolinaWiktoria:
Thanks I've had some problems with tests this solution with `reduce()` . So I had to fix it with `for loop` :)

PR #1040, opened 2023-08-14
solution

mariuszgit:


src/groupByKey.ts
```diff
@@ -1,7 +1,18 @@
+/* eslint-disable no-restricted-syntax */
 type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items, key): GroupsMap<T> {
```
darokrk, line 6: you can even strict type function parameters by using `(items: T[], key: keyof T): GroupsMap<T>`

src/groupByKey.ts
```diff
@@ -1,7 +1,18 @@
...
+  const result: GroupsMap<T> = {};
+
```
darokrk, line 8: you use variable to store object key in a more readable way, like:

`const itemKey: string = String(item[key]);`

then we avoid using double brackets when calling it directly from the object:

```
if (!result[itemKey]) {
  result[itemKey] = [item];
} else {
  result[itemKey].push(item);
}
```

so it will be more readable :)

darokrk (changes requested):
Nicely done, few small mentions added :)

PR #1044, opened 2023-08-16
solution

Sancz0pansa:


src/groupByKey.ts
mvjl000, line 1: Type Assertion should be used only if you are confident about the type or else it will lead to errors in the program. It's not an evil thing - it has some useCases, but I think it shouldn't be in here. If you want to dive deeper, look for `as keyword` in TS docs.

`as string` is not necessary in this solution. I wrote more about it in one of your other tasks review.

mvjl000 (changes requested):

choeqq (approved):
Looks good to me ✅

PR #1049, opened 2023-08-16
add sort method

Meg-Sowka:


domalewski (approved):
good use of generic types

choeqq (approved):
Looks good to me ✅

PR #1055, opened 2023-08-16
Solution

klaudia-klis:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(
+  items: object[],
+  key: keyof object,
+): GroupsMap<object> {
```
choeqq, line 8:
```suggestion
export function groupByKey<T>(
  items: T[],
  key: keyof T,
): GroupsMap<T> {
```
it would be better to use generic here

choeqq (changes requested):

choeqq (approved):
Good job ✅

PR #1069, opened 2023-08-20
add task solution

arturwisnia:


src/groupByKey.ts
```diff
@@ -1,7 +1,22 @@
-type GroupsMap<T> = {
-  [key: string]: T[];
-};
+export function groupByKey<T>(items: T[], key: keyof T): object {
```
choeqq, line 1: the return type will be `GroupsMap<T>`

src/groupByKey.ts
```diff
@@ -1,7 +1,22 @@
...
-export function groupByKey(items, key) {
-  // write code here;
+    if (newGroup) {
+      newGroup.push(item);
+    } else {
+      map.set(keyValue, [item]);
+    }
+
+    return map;
+  }, new Map<any, T[]>());
```
choeqq, line 13: you can make it work using only reduce loop

src/groupByKey.ts
```diff
@@ -1,7 +1,22 @@
...
+    const keyValue = item[key];
+    const newGroup = map.get(keyValue);
 
-export function groupByKey(items, key) {
-  // write code here;
+    if (newGroup) {
+      newGroup.push(item);
+    } else {
+      map.set(keyValue, [item]);
+    }
```
choeqq, line 10:
```suggestion
    if (!map[keyValue]) {
      map[keyValue] = [];
    } 
    
    map[keyValue].push(item)
```

src/groupByKey.ts
choeqq, line 13:
```suggestion
  }, {});
```

choeqq (changes requested):

choeqq (approved):
Looks good to me ✅

PR #1075, opened 2023-08-21
solution

Capoverde:
I stuffed this code with the --no-verify option, because Lint doesn't work for me.
Let me know if it works!


choeqq (approved):
Looks good to me ✅

PR #1079, opened 2023-08-25
group by key

Orchibald:


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+    const keyValue = item[key] as string;
+
+    if (!grouped[keyValue]) {
+      grouped[keyValue] = [];
+    }
+
+    grouped[keyValue].push(item);
+  }
+
+  return grouped;
```
VitaliyBondarenko1982, line 19: try to not disable eslint.
use reduce method instead for loop for it

VitaliyBondarenko1982 (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((grouped, item) => {
+    const keyValue = item[key] as string;
+    const newGrouped = grouped;
```
etojeDenys, line 8: why do we need this var if we can use `grouped` directly

etojeDenys (changes requested):

src/groupByKey.ts
Orchibald: because lint blocks it

Orchibald: image

Orchibald: i have used '--no-verify' for commit new changes

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+  return items.reduce((grouped, item) => {
+    const keyValue = item[key] as string;
+
+    if (!grouped[keyValue]) {
+      grouped[keyValue] = [];
+    }
+
+    grouped[keyValue].push(item);
+
+    return { ...grouped };
```
VitaliyBondarenko1982, line 15:
```suggestion
    return grouped;
```
redundant to destructure grouped.

VitaliyBondarenko1982 (approved):
Well done.

PR #1083, opened 2023-08-29
add task solution

Wesses:
My typescript compiler shows errors, but I passed all the gest tests. Please tell me what is wrong with my code.

зображення
зображення


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const sortedItemsGroup: GroupsMap<T> = {};
+
+  items.reduce((acc: GroupsMap<T>, el: T) => {
```
roman-mirzoian, line 8: Are you sure you need a `reduce` here?
You have created an `sortedItemsGroup` object where you will write information and it is enough for you to walk through the array using `forEach`.

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+    if (!(el[key] in acc)) {
```
roman-mirzoian, line 9: So that there is no mistake, use a more understandable check:
```suggestion
    const objectKey = el[key];
    if (!sortedItemsGroup[objectKey]) { // sortedItemsGroup should be used provided that you will not use reduce
```

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

roman-mirzoian (approved):
Well done 👍

PR #1084, opened 2023-08-30
Solution

oksana-vas:


IvanFesenko (approved):
Good job. For practice try implement this functionality using `Array.reduce()`

oksana-vas:
Thank you for the comment! It was interesting to make the solution using Array.reduce() :)

PR #1086, opened 2023-09-01
Implemented groupping by key logic

davinnchii:


src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], searchedKey: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+  const searchedKeys = new Set(items.map((value) => value[searchedKey]));
```
polosanya, line 7: There is no need to create a Set from array of keys. It creates unnecessary computational complexity.

src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
...
+
+  searchedKeys.forEach((key) => {
+    result[key] = items.filter((item) => item[searchedKey] === key);
```
polosanya, line 10: The usage of the filter method inside a forEach loop can lead to performance issues because it iterates over the entire array for each key. Consider using a single loop for efficiency.

polosanya (changes requested):
Almost done! Let's improve your code a bit)

Denys14Samoilenko (approved):
Nice work 👍🏻

dffUqp (approved):
🚀

PR #1089, opened 2023-09-03
add solution

AndrMar1939:
I feel something wrong in my solution...

### DIFF
--- a/src/groupByKey.ts
+++ b/src/groupByKey.ts
@@ -2,6 +2,27 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+interface Item {
+  [key: string]: any;
+}
+
+export function groupByKey(items :Item[], key: string)
+  :GroupsMap<Item> {
+  const groupedByKey = items.reduce((acc :GroupsMap<Item>, current :Item) => {
+    const newKey :string = current[key];
+
+    if (acc[newKey]) {
+      acc[newKey].push(current);
+
+      return acc;
+    }
+
+    acc[newKey] = [];
+
+    acc[newKey].push(current);
+
+    return acc;
+  }, {});
+
+  return groupedByKey;
 }


src/groupByKey.ts
```diff
@@ -2,6 +2,27 @@ type GroupsMap<T> = {
...
+      acc[newKey].push(current);
+
+      return acc;
+    }
+
+    acc[newKey] = [];
+
+    acc[newKey].push(current);
+
+    return acc;
```
voievodik, line 24:
```
    if (!acc[newKey]) {
      acc[newKey] = [];
    }

    acc[newKey].push(current);

    return acc;
```

voievodik (approved):
Seems interesting and it works :) 
I suggest how you can simplify array pushing

src/groupByKey.ts
polosanya: Good point, let's fix it)

src/groupByKey.ts
```diff
@@ -2,6 +2,27 @@ type GroupsMap<T> = {
...
-export function groupByKey(items, key) {
-  // write code here;
+interface Item {
+  [key: string]: any;
+}
+
+export function groupByKey(items :Item[], key: string)
+  :GroupsMap<Item> {
+  const groupedByKey = items.reduce((acc :GroupsMap<Item>, current :Item) => {
+    const newKey :string = current[key];
```
polosanya, line 12:
```suggestion
    const newKey: string = current[key];
```

src/groupByKey.ts
```diff
@@ -2,6 +2,27 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+interface Item {
+  [key: string]: any;
+}
+
+export function groupByKey(items :Item[], key: string)
+  :GroupsMap<Item> {
+  const groupedByKey = items.reduce((acc :GroupsMap<Item>, current :Item) => {
```
polosanya, line 11:
```suggestion
  const groupedByKey = items.reduce((acc: GroupsMap<Item>, current: Item) => {
```
Fix all cases

polosanya (changes requested):
Almost perfect!

src/groupByKey.ts
```diff
@@ -2,6 +2,23 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+interface Item {
+  [key: string]: any;
+}
+
+export function groupByKey(items: Item[], key :keyof GroupsMap<Item>):
+GroupsMap<Item> {
```
Moroz-Dmytro, line 10:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
```

Moroz-Dmytro: You can use [generic](https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-types) here

src/groupByKey.ts
```diff
@@ -2,6 +2,23 @@ type GroupsMap<T> = {
...
+  const groupedByKey = items.reduce((acc: GroupsMap<Item>, current: Item) => {
+    const newKey: string = current[key];
```
Moroz-Dmytro, line 12:
```suggestion
    const currentItemGroup = String(current[key]);
```

Moroz-Dmytro (changes requested):

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key :keyof T):
```
dffUqp, line 5:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T):
```

dffUqp (approved):
🚀

PR #1098, opened 2023-09-04
added logic for grouping objects by key

irahos:


Orpheus29 (approved):
GJ!

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
```
Moroz-Dmytro, line 6: `result` is not really good name

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item) => {
+    const itemKey: string = String(item[key]);
```
Moroz-Dmytro, line 9:
```suggestion
    const itemGroup = String(item[key]);
```

Moroz-Dmytro (changes requested):

BudnikOleksii (approved):
Good work 👍

PR #1099, opened 2023-09-04
add task solution

Orpheus29:


polinavafik (approved):
Splendid! 🎀

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const sortedGroups: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const sortValue = item[key] as string;
```
Moroz-Dmytro, line 9:
```suggestion
    const sortValue = String(item[key]);
```

Moroz-Dmytro: Better not to use `as`

Moroz-Dmytro (changes requested):

Moroz-Dmytro (approved):

PR #1101, opened 2023-09-04
added groupByKey solution

NazarVotkanych:


desevoker (approved):
Well done! 🚀

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groups: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyGroup = String(item[key]);
```
desevoker, line 9:
```suggestion
    const groupKey = String(item[key]);
```

roman-mirzoian, line 9: The result of `item[key]` will be a value that is already of type string, so there is no need to call the `String `class there:
```suggestion
    const groupKey = item[key];
```

roman-mirzoian (approved):
Well done, but pay attention to the comment.

PR #1102, opened 2023-09-04
initiall commit: create function groupByKey with TS

maksym-mishchanchuk:


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+  for (let i = 0; i < items.length; i += 1) {
+    const item = items[i];
+    const keyValue = String(item[key]);
+
+    if (!groupedItems[keyValue]) {
+      groupedItems[keyValue] = [];
+    }
+
+    groupedItems[keyValue].push(item);
+  }
```
vserhiichuk, line 17: You can use method forEach
When I did a for loop, TS fixed me

vserhiichuk (approved):
It looks good

src/groupByKey.ts
Moroz-Dmytro: Good point, use `forEach` here

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedItems: GroupsMap<T> = {};
+
+  for (let i = 0; i < items.length; i += 1) {
+    const item = items[i];
+    const keyValue = String(item[key]);
```
Moroz-Dmytro, line 10:
```suggestion
    const itemGroup = String(item[key]);
```

Moroz-Dmytro (changes requested):

dffUqp (approved):
GJ👍

PR #1106, opened 2023-09-04
add groupByKey<T> function

neovendano:


KatOlista (approved):
good job

dffUqp (changes requested):
Here's an sample that you can use to rewrite your code because we're doing a lot of iterations right now.

```js
   if (!groupedItems[value]) {
      groupedItems[value] = [];
    }

    groupedItems[value].push(...);
   ```

PR #1121, opened 2023-09-07
add task solution

viktoria-roik:


alexander-ignatow (approved):
Looks good 👍 

IMO, solution with `reduce` would be more elegant

PR #1124, opened 2023-09-08
task solution was added

Liliya-Kalinichenko:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(
+  items:T[],
+  key: keyof T,
+): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
```
Anastasiia-Svintsova, line 11: Refactor it by using the reduce method. You don't really need to create a variable with an empty array

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (approved):

PR #1132, opened 2023-09-17
add task solution

ydem1:


src/groupByKey.ts
```diff
@@ -2,6 +2,24 @@ type GroupsMap<T> = {
...
+
+    if (obj[keyObj] !== undefined && obj[keyObj].length !== 0) {
+      obj[keyObj].push(item);
+    } else {
+      obj[keyObj] = [];
+      obj[keyObj].push(item);
+    }
+  });
+
+  return obj;
```
VitaliyBodnar, line 24: - don't need to create additional variable `obj `, you can use method reduce for in this case
 - `keyObj` is not a correct name, it's better to rename it to `keyValue`

src/groupByKey.ts
```diff
@@ -2,6 +2,24 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item) => {
+    const keyObj = item[key] as string;
+
+    if (obj[keyObj] !== undefined && obj[keyObj].length !== 0) {
+      obj[keyObj].push(item);
+    } else {
+      obj[keyObj] = [];
+      obj[keyObj].push(item);
+    }
```
VitaliyBodnar, line 21: can be less
```suggestion
    if (!obj[keyObj]?.length) {
       obj[keyObj] = [];
    }
    obj[keyObj].push(item);
```

VitaliyBodnar (changes requested):

VitaliyBondarenko1982 (approved):
Well done

PR #1144, opened 2023-10-04
add group by key

dubyuliia:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+    const itemKey = String(item[key]);
+
+    if (grouped[itemKey]) {
+      grouped[itemKey].push(item);
+    } else {
+      grouped[itemKey] = [item];
+    }
+  });
+
+  return grouped;
```
VitaliyBondarenko1982, line 18: suggest to use reduce method instead forEach.
and immediately return result of reduce.

VitaliyBondarenko1982 (changes requested):
Good job.
Let's make some refactoring.

VitaliyBondarenko1982 (approved):
Well done

PR #1147, opened 2023-10-04
Solution

AllaSerhiienko:


plodnikate (changes requested):
You solution is good, but it's can be simplified using forEach.

maxim2310 (approved):
Good work!!!

PR #1155, opened 2023-10-13
task solution

ydashko:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item) => {
+    const keyForObj = item[key] as string;
+
+    if (!newObject[keyForObj]) {
+      newObject[keyForObj] = [];
+    }
+
+    newObject[keyForObj].push(item);
+  });
```
etojeDenys, line 16: you can rewrite it all using the reduce, it will simplify the logic and improve the readability

etojeDenys (approved):
good job

PR #1158, opened 2023-10-16
group by key

whiovt:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  const returnedObject: GroupsMap<T> = {};
+
+  items.forEach((item: T): void => {
+    const currentKey = item[key] as string;
```
anastasiia-tilikina, line 12: using `as` is not a good practice in ts. try to avoid it.

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  const returnedObject: GroupsMap<T> = {};
+
+  items.forEach((item: T): void => {
```
anastasiia-tilikina, line 11:
```suggestion
  items.forEach((item) => {
```

you don't need typing here. TS is already aware what type it is

anastasiia-tilikina (changes requested):
let's improve your code a bit

PR #1160, opened 2023-10-17
add task solution

atrofymovych:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+    const newRes = result;
+
+    if (!result[keyValue]) {
+      newRes[keyValue] = [];
+    }
+
+    result[keyValue].push(item);
+
+    return result;
+  }, {} as GroupsMap<T>);
```
etojeDenys, line 17:
```suggestion
  return items.reduce<GroupsMap<T>>((result, item) => {
    const keyValue = item[key] as string;
    const newRes = result;

    if (!result[keyValue]) {
      newRes[keyValue] = [];
    }

    result[keyValue].push(item);

    return result;
  }, {});
```

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((result, item) => {
+    const keyValue = item[key] as string;
+    const newRes = result;
```
etojeDenys, line 8: it's the same object, so this var looks redundant

etojeDenys (changes requested):

PR #1161, opened 2023-10-17
ts-group-by-key task implementation

yulabakun:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: Array<T>, key: keyof T): GroupsMap<T> {
+  const grouppedData: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyToGroupBy = String(item[key]);
```
vlevik, line 9: it's a good solution for this task, but it would be great to restrict not primitive types to be value of `T` object. for example, `String({ key: "value" })` will be `'[object Object]'`

vlevik (approved):

vlevik (approved):
good job! 🔥

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: Array<T>, key: keyof T): GroupsMap<T> {
+  const grouppedData: GroupsMap<T> = {};
+
+  items.forEach((item) => {
```
StasSokolov, line 8: i think will be better use `reduce` for this case without `grouppedData` variable

StasSokolov (approved):
Looks good to me, i'm approving it, check only one comment for improve your code!

PR #1163, opened 2023-10-17
solution

vlevik:


YevhenKharko (approved):
Good job!

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+      if (acc[groupBy]) {
+        acc[groupBy].push(item);
+      } else {
+        acc[groupBy] = [item];
+      }
+
+      return acc;
+    }, {});
+
+  return groupedItems;
```
StasSokolov, line 21:
```suggestion
  return items
    .reduce((acc: GroupsMap<T>, item: T) => {
      const groupBy = item[key];

      if (acc[groupBy]) {
        acc[groupBy].push(item);
      } else {
        acc[groupBy] = [item];
      }

      return acc;
    }, {});
```

can be more simplify

StasSokolov (approved):
Looks good to me, i'm approving it!

PR #1164, opened 2023-10-17
implemented task solution

YevhenKharko:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
```
StasSokolov, line 8: i think will be better use `.reduce` in this case without `result` variable

StasSokolov (approved):
Looks good to me, i'm approving it, check only one comment from me for improve your code a bit more!

PR #1166, opened 2023-10-17
solution

OlegPopovych:
Hello!

### DIFF
--- a/src/groupByKey.ts
+++ b/src/groupByKey.ts
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof string): GroupsMap<T> {
+  const result = {};
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const item of items) {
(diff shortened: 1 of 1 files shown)


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof string): GroupsMap<T> {
+  const result = {};
+
+  // eslint-disable-next-line no-restricted-syntax
```
tilugdmytro, line 8: Very smart! 😃

tilugdmytro (approved):
Good job!

src/groupByKey.ts
StasSokolov, line 8: let's remove eslint disable and only change `for of` to `.reduce()`

StasSokolov (changes requested):
Great job, only one minor comment from me!

src/groupByKey.ts
OlegPopovych: Done.

StasSokolov (approved):
Awesome job, i'm approving it🔥🔥

PR #1167, opened 2023-10-17
add task solution

tilugdmytro:


ElenaGrasovskaya (approved):
A great solution, and I bet it's fast!

tilugdmytro:
@ElenaGrasovskaya  I wish it was that simple 😄

src/groupByKey.ts
```diff
@@ -2,6 +2,23 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T, K extends keyof T>(
+  items: T[],
+  key: K,
+): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.map((item) => {
```
StasSokolov, line 11: will be better use `.reduce()` in this case without redundant variable `grouped`

StasSokolov (approved):
Great job, i'm approving it!

PR #1168, opened 2023-10-17
I'm not quite sure how, but it works

ElenaGrasovskaya:


StasSokolov (approved):
Great job, i'm approving it, good solution but will be more informative and simplify if you use `.reduce()`

PR #1175, opened 2023-10-22
add group  by key solution

Kuper18:
Hi, here is my solution

### DIFF
--- a/src/groupByKey.ts
+++ b/src/groupByKey.ts
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey
+<T extends Record<string, any>>(items: T[], key: string): GroupsMap<T> {
+  const object = items.reduce((acc, curr) => {
+    const filter = items.filter((item) => item[key] === curr[key]);
+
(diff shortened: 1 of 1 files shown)


mifyy (approved):

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey
+<T extends Record<string, any>>(items: T[], key: string): GroupsMap<T> {
```
vasylmiskiv, line 6: You can set type directly without extenstion
Also you can set key as a key of the object T
```suggestion
<T>(items: T[], key: keyof T): GroupsMap<T> {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+  const object = items.reduce((acc, curr) => {
+    const filter = items.filter((item) => item[key] === curr[key]);
```
vasylmiskiv, line 8: Would be better to rename it to `groupedItems` instead of `filter`

vasylmiskiv (approved):
Well done 🚀

PR #1177, opened 2023-10-24
solution

BiletskyiVolodymyr:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const grouped: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const value = item[key] as string;
+
+    if (grouped[value]) {
```
VitaliyBodnar, line 11: Make sense to use ternary operator here as it will make a code shorter

VitaliyBodnar (changes requested):

etojeDenys (approved):
awesome

PR #1181, opened 2023-10-25
Solution

NVlasenko:


nazarmatsevych (changes requested):
Good work, code looks nice but let's fix lint errors
image

etojeDenys (approved):
interesting solution

PR #1182, opened 2023-10-26
implementation

Daniilart01:
I have a question about generics. I want to set type in this line `const groupNames: Set<string> = new Set([...(items.map((obj: T) => String(obj[key])))]);` automatically si i don`t need to convert value to string here: `items.map((obj: T) => String(obj[key]))` an here `String(item[key]) === group`. How can i improve my code?

### DIFF
--- a/src/groupByKey.ts
+++ b/src/groupByKey.ts
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T extends Object>(items: T[],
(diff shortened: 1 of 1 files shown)


etojeDenys (approved):
good job

etojeDenys:
> I have a question about generics. I want to set type in this line `const groupNames: Set<string> = new Set([...(items.map((obj: T) => String(obj[key])))]);` automatically si i don`t need to convert value to string here: `items.map((obj: T) => String(obj[key]))`an here`String(item[key]) === group`. How can i improve my code?

feel free to ask the questions in the chat

PR #1196, opened 2023-11-04
Solution

ndv20:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+  items.forEach((el) => {
+    const itemKey = String(el[key]);
+
+    if (!result[itemKey]) {
+      result[itemKey] = [];
+    }
+    result[itemKey].push(el);
+  });
+
+  return result;
```
SanyaBratashchuk, line 17: It can be simplified via reduce method

SanyaBratashchuk (approved):

PR #1197, opened 2023-11-04
Solution

YevgeniySavchuk:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  // eslint-disable-next-line no-restricted-syntax
```
VitaliyBodnar, line 8: try to avoid using eslint disable, cause it's a bad practice

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+  for (const item of items) {
+    if (result[item[key]]) {
+      result[item[key]].push(item);
+    } else {
+      result[item[key]] = [item];
+    }
```
VitaliyBodnar, line 14: make sense to use ternary operator here as it will make a code shorter

VitaliyBodnar (changes requested):

src/groupByKey.ts
YevgeniySavchuk: eslint was not OK with for of loop

src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
...
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((prev, curr) => {
+    const itemKey = curr[key] as string;
+
+    return {
+      ...prev,
+      [itemKey]: prev[itemKey]
+        ? [...prev[itemKey], curr]
+        : [curr],
+    };
```
VitaliyBodnar, line 14: let's make it simpler 
```suggestion
    prev[itemKey] 
       ? prev[itemKey].push(curr)
       : prev[itemKey] = [curr] 
    return prev;
```

VitaliyBodnar (changes requested):

src/groupByKey.ts
YevgeniySavchuk: This way does not work:(
image

YevgeniySavchuk: but in VS Code on js it works!

VitaliyBodnar: Okay, let's leave the previous version

VitaliyBodnar (approved):

PR #1207, opened 2023-11-14
Add task solution

SerhiiUnhurian:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyValue = item[key] as string;
+
+    if (!result[keyValue]) {
+      result[keyValue] = [item];
+    } else {
+      result[keyValue].push(item);
+    }
```
VitaliyBodnar, line 15: make sense to use ternary operator here as it will make a code shorter

VitaliyBodnar (changes requested):

src/groupByKey.ts
SerhiiUnhurian: I think this code is quite short
There will be even more code to fix

src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyValue = item[key] ? (item[key] as string) : '';
+
+    return result[keyValue] ? result[keyValue].push(item)
+      : (result[keyValue] = [item]);
+  });
```
VitaliyBodnar, line 13:
```suggestion
    return result[keyValue] 
      ? result[keyValue].push(item)
      : result[keyValue] = [item];
  });
```

src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyValue = item[key] ? (item[key] as string) : '';
```
VitaliyBodnar, line 9:
```suggestion
    const keyValue = item[key] || '';
```

VitaliyBodnar (changes requested):

src/groupByKey.ts
SerhiiUnhurian: we have a problem here 
Screenshot 2023-11-14 at 15 08 56
Screenshot 2023-11-14 at 15 08 49

VitaliyBodnar: okay, let's try this one 
`const keyValue = item[key] as string || ''`;

VitaliyBodnar (approved):

PR #1212, opened 2023-11-18
solution

ArtemSydun:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
```
polosanya, line 6: Would be great to use more meaningful names, better specify what exactly the result is

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+
+  items.forEach((item: T) => {
+    const value = item[key] as string;
```
polosanya, line 9: Same with value

polosanya (approved):
Good job! 👍

PR #1213, opened 2023-11-19
solution

vinogradova8:


src/groupByKey.ts
```diff
@@ -1,7 +1,30 @@
...
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const arrOfValues: string[] | number[] = [];
+  const result: GroupsMap<T> = {};
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const item of items) {
+    if (!arrOfValues.includes(item[key])) {
+      arrOfValues.push(item[key]);
+    }
+  }
```
anastasiiavorobiova, line 15:
```suggestion

```
Also, you can simplify the logic a bit, work with a result, and use just one cycle here

anastasiiavorobiova (changes requested):
Great work! Added a few comments to fix

src/groupByKey.ts
```diff
@@ -1,7 +1,30 @@
...
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const arrOfValues: string[] | number[] = [];
+  const result: GroupsMap<T> = {};
+
+  // eslint-disable-next-line no-restricted-syntax
```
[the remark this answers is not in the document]
vinogradova8: I have no idea how to avoid using loop hear ( 8:3  error  iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations  no-restricted-syntax)

src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
...
+    const value: string | number = item[key];
+
+    if (result[value] !== undefined) {
+      result[value].push(item);
+    }
+
+    if (result[value] === undefined) {
+      result[value] = [];
+      result[value].push(item);
+    }
```
nazarmatsevych, line 19:
```suggestion
    if (!result[value]) {
      result[value] = [];
    }
    result[value].push(item);
```

we can simplify it a bit

src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const item of items) {
```
nazarmatsevych, line 9: You can use array methods instead of for loop, forEach for example

nazarmatsevych (changes requested):
Almost done, please see comment below :)

etojeDenys (approved):
good job

PR #1224, opened 2023-11-20
Solution

alina-kabanets:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    const value = String(item[key]);
+
+    if (value in result) {
+      result[value].push(item);
+    } else {
+      result[value] = [item];
+    }
```
nazarmatsevych, line 15:
```suggestion
    if (!result[value]) {
      result[value] = [];
    }
      result[value].push(item);
    
```

we can simplify it a bit and remove else block

nazarmatsevych (approved):
Great work, left minor comment :)

PR #1230, opened 2023-11-23
add task solution

yuliiababych:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedObj: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    const keyValue = item[key] as string;
```
TarasHoliuk, line 9: In this case, it works as expected, but note that a value inside an object might be anything (not only string). So it's dangerous - use type casting in such cases. Better do the following:

```suggestion
    const keyValue = String(item[key]);
```
Transform it into a string instead of saying TS that you sure it's a string :)

TarasHoliuk (approved):
Approved :+1: With one small, but important comment - read it, please

PR #1232, opened 2023-11-23
solution

YuriiHlushenko:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T[]): GroupsMap<T> {
+  const keys: string| number[] = [];
+
+  items.map((item) => (keys.includes(item[key])
+    ? null
+    : keys.push(item[key])));
```
etojeDenys, line 10:
```suggestion
  items.forEach((item) => (!keys.includes(item[key]) && keys.push(item[key])));
```
better to use forEach 'vause you don't need the array that map returns

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+
+  const result = {};
+
+  keys.map((k) => (result[k] = items.filter((el) => el[key] === k)));
```
etojeDenys, line 14: same here, better to  use the reduce

etojeDenys (changes requested):

src/groupByKey.ts
YuriiHlushenko: I can't think of how to use reduce here

lerastarynets: @YuriiHlushenko try to rewrite it with reduce when u have time)

lerastarynets (approved):
lgtm 🚀

PR #1236, opened 2023-11-24
Solution

Patrykmclaren14:


mbruhler (changes requested):
When I run your function like:
```
const items = [
  { id: 1, name: 'John' },
  { id: 2, name: 'Jane' },
  { id: 3, name: 'John' },
];

const grouped = groupByKey(items, 'name');
```
It throws ts error
`Type '{ id: number; name: string; }' is not assignable to type 'Item'`

So your code doesn't work for objects at all. Please fix it so one would group the array of objects too.

I think it should be `export function groupByKey<T>(` but please check it.

PR #1242, opened 2023-11-29
generic types

BOHDANLIUYUI:


src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((grouped: GroupsMap<T>, item: T) => {
+    const keyV = item[key] as string;
```
TarasHoliuk, line 7:
```suggestion
    const keyV = String(item[key]);
```
Since theoretically value of the object may be anything it's better to transform it into string instead of using type casting - FYI

src/groupByKey.ts
```diff
@@ -2,6 +2,13 @@ type GroupsMap<T> = {
...
+
+    return {
+      ...grouped,
```
TarasHoliuk, line 10: you might mutate `grouped` object and return in instead of making copy - just FYI

TarasHoliuk (approved):
Approved with 2 comments (just FYI)

PR #1245, opened 2023-11-30
v1

DVasylenko18:


src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
+export function groupByKey<T>(items: T[], key: keyof T):GroupsMap<T> {
   // write code here;
+  const result: GroupsMap<T> = {};
+  const distinct = [...items.map((item:T) => `${item[key]}`)];
+
+  for (let i = 0; i < distinct.length; i += 1) {
```
volodymyr-soltys97, line 10: Use a forEach method here instead of for loop

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1247, opened 2023-12-02
solution

AnastasiiaKriuchkova:


src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupItems: GroupsMap<T> = {};
+
+  items.forEach((item) => {
```
VitaliyBondarenko1982, line 8: suggest to use reduce method instead forEach and immediately return result of reduce
```
return items.reduce((groupItems: GroupsMap<T>, item) => {
   do something
})
```

src/groupByKey.ts
```diff
@@ -2,6 +2,14 @@ type GroupsMap<T> = {
...
+    const itemValue = String(item[key]);
+
+    groupItems[itemValue] = items.filter((obj) => obj[key] === item[key]);
```
VitaliyBondarenko1982, line 11: it is redundsnt to filter items on each iteration.
you can push to existance value or create empty array in case if no exists itemValue key in groupItems.
something like that
```
groupItems[itemValue] = acc[itemValue] || [];
groupItems[itemValue].push(item);
```

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some refactoring.
Please check comments.

VitaliyBondarenko1982 (approved):
Well done

PR #1257, opened 2023-12-12
Tests passed

momos1703:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+    const valueStr = String(item[key]);
+
+    if (!result[valueStr]) {
+      result[valueStr] = [];
+    }
+
+    result[valueStr].push(item);
+  });
+
+  return result;
```
VitaliyBodnar, line 21: - consider to using method reduce here instead of forEach as it will make your code shorter

VitaliyBodnar (changes requested):

src/groupByKey.ts
momos1703: I've try using reduce but had problem with types...

momos1703: And it wasn't much shorter.

nazarmatsevych (approved):
Great work!

PR #1266, opened 2023-12-18
solution

VladimirKolesnikov:
I understand that using of type assertion (str 6, 14) is not perfect but I don't know how to avoid it.
Anyway we can't reveal type of T-object values before compilation.


lerastarynets (approved):
https://opportunity.atlassian.net/browse/ETA-8230

PR #1277, opened 2023-12-27
Solution

InnaSh23:


src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
+  return items.reduce<GroupsMap<T>>((acc, item) => {
+    const keyValue = item[key] as unknown as string;
```
nazarmatsevych, line 7:
```suggestion
    const keyValue = item[key].toString();
```

I think we can do like this

nazarmatsevych (approved):
Looks great 👍 
Left a minor comment

PR #1287, opened 2024-01-07
Solved task

vashivan:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const keyValue = item[key] as string;
```
TarasHoliuk, line 9: FYI: it's better to make it a string instead of using typecasting:

```suggestion
    const keyValue = String(item[key]);
```

TarasHoliuk (approved):
GJ - Approved :heavy_check_mark: 
Added 1 comment just FYI

vashivan:
Got it, thank you 🙏🏻

PR #1298, opened 2024-01-12
add task solution

artem8746:


nabokok (approved):
Great work! Keep it up

src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
+  // eslint-disable-next-line no-restricted-syntax
+  for (const item of items) {
+    const groupField = String(item[key]);
+
+    if (!groupedObjects[groupField]) {
+      groupedObjects[groupField] = [];
+    }
+
+    groupedObjects[groupField].push(item);
+  }
```
DanilWeda, line 17: Try to use the array method which returns an object for a solution.
So, you do not need object results and can write the method shortly.

DanilWeda (changes requested):
Good job! 
Check the improvements comment.
Thanks!

src/groupByKey.ts
artem8746: but why? we use it just here

DanilWeda (approved):
Excellent! 
Approved!
Many thanks!

PR #1300, opened 2024-01-13
add task solution

rvoshchylo:


vSainiuk (approved):
Great!

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+
+  items.forEach((obj: R) => {
+    const keyInObj = obj[key] as string;
+
+    if ((keyInObj) in sortedObjects) {
+      sortedObjects[keyInObj].push(obj);
+    } else {
+      sortedObjects[keyInObj] = [obj];
+    }
+  });
```
DanilWeda, line 16: You can use another method which returns an object. (**recommendation***)

DanilWeda (approved):
Good. 
Recommended use for this task another array method.
Approved, but please try to think about realisation with different method.
Many thanks!

PR #1311, opened 2024-01-20
task solution

BeginerAlex:


VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some  refactoring. use reduce method instead forEach.

src/groupByKey.js
```diff
@@ -0,0 +1,28 @@
...
+            s = arguments[i];
+            for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p))
+                t[p] = s[p];
+        }
+        return t;
+    };
+    return __assign.apply(this, arguments);
+};
+exports.__esModule = true;
+exports.groupByKey = void 0;
```
volodymyr-soltys97, line 14: Remove it
```suggestion
```

src/groupByKey.js
```diff
@@ -0,0 +1,28 @@
...
+        if (itemKey in updatedGroupedItems) {
+            updatedGroupedItems[itemKey].push(item);
+        }
+        else {
+            updatedGroupedItems[itemKey] = [item];
+        }
+        return updatedGroupedItems;
+    }, {});
+}
+exports.groupByKey = groupByKey;
```
volodymyr-soltys97, line 28: Redundant, you need to write code only in `groupByKey.ts` file

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/groupByKey.js
BeginerAlex: I got it, but I have some problems with the compiler and the versioning node. To compile my .ts file I need to delete my .js file and create a new compiler file again. 

(I specify the path to my .ts file in tsconfig.json and always run the command "tsc -p tsconfig.json --watch" but it doesn't work. It's probably a problem with my .js file because of that)

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((groupedItems: GroupsMap<T>, item: T) => {
+    const itemKey = String(item[key]);
+    const updatedGroupedItems = { ...groupedItems };
```
etojeDenys, line 8: why do we need to create the copy here, looks redundant

etojeDenys (approved):
good job

PR #1312, opened 2024-01-20
solution

Mariana-VV:


src/groupByKey.ts
```diff
@@ -2,6 +2,25 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: [], key: string):
```
etojeDenys, line 5:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T>
```

etojeDenys (changes requested):
add better types for function

etojeDenys (approved):
good job

PR #1314, opened 2024-01-21
solution group by key

TVChulkova:


src/groupByKey.ts
```diff
@@ -2,6 +2,27 @@ type GroupsMap<T> = {
...
+interface Item {
+  'id': number,
+  'color': string,
+  'country': string,
+}
+
+export function groupByKey(
+  items: Item[],
+  key: keyof Item,
+): GroupsMap<Item> {
```
etojeDenys, line 14:
```suggestion
export function groupByKey<T>(
  items: T[],
  key: keyof T,
): GroupsMap<T> {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,27 @@ type GroupsMap<T> = {
...
+
+  items.map((item: Item) => {
+    if (!(Object.hasOwn(result, item[key]))) {
+      result[item[key] as keyof typeof result] = [
+        ...items.filter((itemFilter: Item) => itemFilter[key] === item[key]),
+      ];
+    }
+
+    return undefined;
+  });
```
etojeDenys, line 25: you can try to simplify it using the reduce

etojeDenys (changes requested):

src/groupByKey.ts
TVChulkova: I've changed my solution. Please watch it once more. Thx

etojeDenys (approved):
awesome

PR #1318, opened 2024-01-22
sol

Patryk-Buczkowski:


src/groupByKey.ts
```diff
@@ -2,6 +2,40 @@ type GroupsMap<T> = {
...
+export function groupByKey(items: [], key: string): GroupsMap<typeof items> {
+  const result: GroupsMap<typeof items> = {};
+
+  items.map((item) => {
+    result[item[key]] = [];
+
+    return result;
+  });
+
+  items.map((item) => {
```
volodymyr-soltys97, line 14: You don't need to duplicate the same method for the array twice, you need to write the logic inside one method, think how it can be fixed

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

nazarmatsevych (approved):
Good work!

PR #1322, opened 2024-01-23
add solution

ssaviak:


vSainiuk (approved):
Marvelous 💫

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const resObj: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
+    const keyOfItem = item[key] as string;
```
DanilWeda, line 9:
```suggestion
    const keyOfItem = String(item[key]);
```

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const resObj: GroupsMap<T> = {};
+
+  items.forEach((item: T) => {
```
DanilWeda, line 8: You can use reduce instead forEach. In this case you don't need to create resObj on 6 line.

DanilWeda (approved):
Good job! 
Check comments for future solutions.

src/groupByKey.ts
ssaviak: Thanks. i was looking for solution like this

PR #1329, opened 2024-01-30
add task solution

YevhenProtasov:


etojeDenys (approved):
good job, you can try to rewrite it using the reduce

PR #1338, opened 2024-02-05
add task solution

yuriiyepikhov:


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
+export function groupByKey<R>(items: R[], key: keyof R): GroupsMap<R> {
   // write code here;
+  const result: GroupsMap<R> = {};
+
+  items.forEach((item: R) => {
+    const value = item[key];
+
+    if (`${value}` in result) {
```
etojeDenys, line 12: ``${value}``  you can move this to some var

etojeDenys (approved):
good job, you can try to rewrite it using the reduce method

PR #1343, opened 2024-02-07
add task solution

legogoUA:


etojeDenys (approved):
looks good, you can try to rewrite it using the reduce method

PR #1350, opened 2024-02-13
solved

DanielLinker17:


etojeDenys (approved):
lgtm, you can simplify it using reduce

PR #1353, opened 2024-02-15
add task solution

ArsenSkandalenko:


etojeDenys (changes requested):
LGTM, but you can try to simplify it with the reduce method

etojeDenys (approved):

PR #1355, opened 2024-02-17
grouped

edkido:


etojeDenys (approved):
lgtm, but you can try to simplify it using the reduce

PR #1365, opened 2024-02-25
added solution ts group by key

nk8bnj:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
```
polosanya, line 5: key is not just a string, let's set the correct type

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+  return items.reduce((groups: GroupsMap<T>, item: T) => {
+    const groupKey = item[key as keyof T] as unknown as string;
```
polosanya, line 7: This part is overcomplicated. Try to simplify it

polosanya (changes requested):
Good job! Let's improve your code a bit)

polosanya (changes requested):
Comments above

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((groups: GroupsMap<T>, item: T) => {
+    const groupKey = String(item[key]);
+
+    const updatedGroups = { ...groups };
```
Moroz-Dmytro, line 9: Why do you need to make a copy every time?

Moroz-Dmytro (approved):

PR #1370, opened 2024-02-28
solution

Arkadiuszx:


yevhenii-pyl (approved):
Good job! 👍 

Btw, here you have implemented something called 'normalization' - it's a widely used technic, this is definitely a technic you should be aware of. 
To read a bit more: https://dev.to/danielpdev/normalize-your-complex-js-objects-21d9

PR #1384, opened 2024-03-06
awesome solution

VitaliiNez:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const obj: GroupsMap<T> = {};
+
+  items.forEach((el) => {
+    if (Object.hasOwn(obj, el[key] as string)) {
+      obj[el[key] as string].push(el);
+    } else {
+      obj[el[key] as string] = [el];
+    }
```
etojeDenys, line 13:
```suggestion
  items.forEach((el) => {
    const betterToDoLikeThat = el[key] as string;
    if (Object.hasOwn(obj, betterToDoLikeThat)) {
      obj[betterToDoLikeThat].push(el);
    } else {
      obj[betterToDoLikeThat] = [el];
    }
```

etojeDenys (approved):
you can try to rewrite it using the reduce method

PR #1388, opened 2024-03-08
Solution

Aksahir:


src/groupByKey.ts
```diff
@@ -1,7 +1,22 @@
...
-export function groupByKey(items, key) {
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
   // write code here;
+  const grouped: GroupsMap<T> = {};
+
+  items.forEach((item) => {
+    const value: T[keyof T] = item[key];
+
+    if (!grouped[String(value)]) {
+      grouped[String(value)] = [];
```
anastasiiavorobiova, line 15: You can convert the value to a string once and use it further in the code to avoid repeating conversions

anastasiiavorobiova (approved):
Good work!

PR #1397, opened 2024-03-12
TaskComplete

TideTwo:
The task itself was easy, but I don't think I will be able to keep up with how convoluting the typing can get, I dread seeing it in an actual complex project instead of just an example task 😅


TideTwo:
Again issue with Node, prettier and tests. Local check pass proof:
image

src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
```
choeqq, line 5:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
```

choeqq (changes requested):

TideTwo:
Changing type of key parameter makes sense, but I don't see changing ` as keyof GroupsMap<T>` to ` as string` to benefit anything

choeqq (approved):

PR #1401, opened 2024-03-13
solution

olduma:


ViktoriiaRepo (approved):
Good job!

Moroz-Dmytro (approved):
Looks good 👍 
But maybe would be better to use `reduce` here

PR #1403, opened 2024-03-14
created TS function groupByKey

YarynaPuhach:


src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  return items.reduce((acc, item) => {
+    const keyValue = item[key] as unknown as string;
```
Davwidos, line 7: In this example it  works but i think T should extend somethig that has only string and numbers as values so it is impossible to use groupByKey with wrong objects T.

Davwidos (changes requested):
function groupByKey should be impossible to use for objects `T` that have in `T[key] ` something that can't be key in object

choeqq (approved):

PR #1417, opened 2024-03-23
add solution and update test node version

IhorRome:


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
+export function groupByKey<T>(items: T[], key: string): GroupsMap<T> {
```
etojeDenys, line 5:
```suggestion
export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
```

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1426, opened 2024-03-27
version 1.1

Maksym-Atanasov:


src/groupByKey.ts
```diff
@@ -1,7 +1,20 @@
+/* eslint-disable max-len */
```
anastasiia-tilikina, line 1: Why do you need to disable linter here?

anastasiia-tilikina (approved):
GJ

PR #1434, opened 2024-04-02
solution

maksym-kostetskyi:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const arrOfKeys: any[] = [];
```
Viktor-Kost, line 6: redundant
```suggestion
```

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+    if (!arrOfKeys.includes(items[i][key])) {
+      arrOfKeys.push(items[i][key]);
+    }
+  }
+
+  for (let i = 0; i < arrOfKeys.length; i++) {
+    result[arrOfKeys[i]] = items.filter(
+      (element) => element[key] === arrOfKeys[i],
+    );
+  }
```
Viktor-Kost, line 19: you can simplify this logic to avoid having two loops
```suggestion
  items.forEach((item) => {
    const keyValue = item[key] as string;

    if (!result[keyValue]) {
      result[keyValue] = [];
    }

    result[keyValue].push(item);
  });
```

Viktor-Kost (changes requested):

loralevitska (approved):
Well done)

PR #1436, opened 2024-04-03
add task solution

dyarechuk:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+    const groupKey: string = item[key];
+
+    if (!result[groupKey]) {
+      result[groupKey] = [];
+    }
+
+    result[groupKey].push(item);
+  }
+
+  return result;
```
DanilWeda, line 21: You can use reduce for it. 
Try to write this variant.

DanilWeda (changes requested):
Good job. 
Check small comments. 
Many thanks)

PR #1437, opened 2024-04-03
work

FoReWwEr:


etojeDenys (approved):
lgtm, you can try to simplify it using the reduce method

PR #1444, opened 2024-04-10
ts_group_by_key

georgiyevv:


anastasiiavorobiova (changes requested):
Great work! Almost done!

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<Item>(
+  items: Array<Item>,
+  key: keyof Item,
```
etojeDenys, line 7:
```suggestion
export function groupByKey<T>(
  items: Array<T>,
  key: keyof T,
```
we usually just name it as `T`

etojeDenys (approved):
lgtm, but better to rewrite it using the reduce method

PR #1454, opened 2024-04-19
task solution

MatheusStefan:


src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T extends object>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  const groupedMap: GroupsMap<T> = {};
+
+  items.map((item) => {
```
EdPirro, line 11: See how you ignored `.map`'s return and used the external object `groupedMap` to store each step's result? This is a clear indicator that you're using the wrong array function!

`.reduce` would be more fitting to this problem

src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
...
+    const keyValue: any = item[key];
+    const keyString = String(keyValue);
+
+    if (Object.prototype.hasOwnProperty.call(groupedMap, keyString)) {
```
EdPirro, line 15: Why not `if (groupedMap.hasOwnProperty(keyString))` or simply `if (groupedMap[keyString])`?

EdPirro (changes requested):
Good job, @MatheusStefan! 👏
I pointed out just a few improvements, please go through them and let me know if you have any questions

src/groupByKey.ts
MatheusStefan: I was studying some new things, but your approach is way simple, thanks!

src/groupByKey.ts
```diff
@@ -2,6 +2,20 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T extends object>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  return items.reduce((groupedMap, item) => {
+    const keyValue: any = item[key];
+    const keyString = String(keyValue);
```
jv-aquino, line 11: nice spoting here of type safety (that is, we don't need to add keyString: string because String function ensures that!)

jv-aquino (approved):
Great job! Concise and absolutely type safe code!

PR #1458, opened 2024-04-21
add solution

reznik-denis:


etojeDenys (approved):
lgtm, but you can simplify it using the reduce method

PR #1461, opened 2024-04-25
adding task solution

eduvinagre:


pedro-ruas (approved):
Noice! Very good use of Generics here

PR #1470, opened 2024-04-28
add solution

iAlfreX:


src/groupByKey.ts
```diff
@@ -2,6 +2,19 @@ type GroupsMap<T> = {
...
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T extends object>(
+  items: T[],
+  key: keyof T,
+): GroupsMap<T> {
+  return items.reduce((groupedData: GroupsMap<T>, item) => {
+    const valueOfKey: unknown = item[key];
```
polosanya, line 10: What's the purpose of making it `unknown`? Let's make sure that it's string

polosanya (changes requested):
Almost done!

DanilWeda (approved):
Good job. 
Approved. 
Thanks)

PR #1476, opened 2024-05-02
add task solution

OleksandraKoshyk:


src/groupByKey.ts
```diff
@@ -2,6 +2,23 @@ type GroupsMap<T> = {
...
+    if (values.indexOf(item[key]) === -1) {
+      values.push(item[key]);
+    }
+  });
+
+  values.forEach((value) => {
+    result[String(value)] = filterBy(items, value);
+  });
+
+  return result;
```
volodymyr-soltys97, line 23: You can simplify your code
```suggestion
  const result: GroupsMap<T> = {};
  
  for (const obj of items) {
    const keyValue = obj[key] as string;

    if (!result[keyValue]) {
      result[keyValue] = [];
    }
    result[keyValue].push(obj);
  }
  
  return result;
```

volodymyr-soltys97 (approved):
Good job 👍

PR #1480, opened 2024-05-09
Task ts group by key

HigorPedroso:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
...
+    const value = item[key] as string;
+
+    if (!result[value]) {
+      result[value] = [];
+    }
+
+    result[value].push(item);
+  });
+
+  return result;
```
EdPirro, line 18: See how you're using an external object in your `.forEach`?  When you want to transform an array into one single value (in this case an object) prefer `.reduce` which eliminates the usage of an external variable

EdPirro (approved):
Great job, @HigorPedroso! Keep it up! 🚀

PR #1482, opened 2024-05-11
GroupByKey

VadymBryvka:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: any = {};
```
vadiimvooo, line 6:
```suggestion
  const result: GroupsMap<T> = {};
```

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+
+  for (const i of items) {
+    if (typeof result[`${i[key]}`] !== 'undefined') {
+      result[i[key]].push(i);
+    } else {
+      result[i[key]] = [];
+      result[i[key]].push(i);
+    }
```
vadiimvooo, line 14:
```suggestion
    const value = String(i[key]);

    if (!result[value]) {
      result[value] = [];
    }
    result[value].push(i);
```

vadiimvooo (changes requested):
Good job. A few changes required.

volodymyr-soltys97 (approved):
Good job! 👍

PR #1487, opened 2024-05-16
solution

YuliiaCat:
https://github.com/YuliiaCat/ts_group_by_key


etojeDenys (approved):
good job, but you can simplify it using the reduce method

PR #1498, opened 2024-05-24
add solution

Alenalenk:


src/groupByKey.ts
```diff
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey(items: [], key: string): GroupsMap<string> {
+  const itemsValuesOfKeys: string[] = [];
```
etojeDenys, line 6: try to simplify this logic without using additional arrays, you can just check `if (!results[item]) {`

etojeDenys (changes requested):

etojeDenys (approved):
lgtm, but you can simplify it using the reduce method

PR #1515, opened 2024-06-06
Solution

yuliaPel:


src/groupByKey.ts
```diff
@@ -2,6 +2,15 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const groupedKey: GroupsMap<T> = {};
+  const values = items
+    .map((item) => item[key])
+    .filter((item, index, arr) => arr.indexOf(item) === index);
```
roman-mirzoian, line 9: For optimization Instead of using map and filter, it is better to use forEach to immediately add elements to the corresponding groups.

roman-mirzoian (approved):
Well done!

PR #1524, opened 2024-06-30
[Feature] added group by key function

Arthur-Tolkachov:
## Description
Added group by key function


volodymyr-soltys97 (approved):
Great work 🔥

PR #1530, opened 2024-08-02
Solution

Teollan:


BudnikOleksii (approved):
Good work 👍 
You can use `reduce` to generate object

PR #1531, opened 2024-08-02
add solution

RimmaBuzuluk:


BudnikOleksii (approved):
Good work 👍 
You can use `reduce` to generate an object

PR #1534, opened 2024-08-03
Solution

Karanelus:
Implemented a function `groupByKey` that takes an array of similar objects 
(`books`, `students`, etc.) and a string `key` that is one of possible item 
keys (`key: keyof T`).

### DIFF
--- a/src/groupByKey.ts
+++ b/src/groupByKey.ts
@@ -2,6 +2,22 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(object: T[], key: keyof T): GroupsMap<T> {
+  const finalObject: GroupsMap<T> = {};
+
+  object.forEach((el) => {
+    const keyValue = el[key];
+
+    if (keyValue !== undefined && keyValue !== null) {
+      const keyString = String(keyValue);
+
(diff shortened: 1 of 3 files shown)


src/groupByKey.ts
```diff
@@ -2,6 +2,29 @@ type GroupsMap<T> = {
...
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(object: T[], key: keyof T): GroupsMap<T> {
+  const finalObject: GroupsMap<T> = {};
+
+  object.forEach((el) => {
+    const keyValue = el[key];
+
+    if (keyValue !== undefined && keyValue !== null) {
```
BogdanMaliuta, line 11:
```suggestion
    if (keyValuel) {
```

src/groupByKey.ts
```diff
@@ -2,6 +2,29 @@ type GroupsMap<T> = {
...
+      const keyString = String(keyValue);
+
+      if (keyString !== '') {
```
BogdanMaliuta, line 14: | think you need to check here your string with trim() 
```suggestion
      if (keyString.trim()) {
```

BogdanMaliuta (changes requested):
Hello, not bad.
But fix my comments and AI

natalia-klonowska (approved):

PR #1536, opened 2024-08-05
Solution

AndGrachov:


src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
```
BudnikOleksii, line 6: `result` is a bad name for a variable

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+  const uniqueValues: any[] = [];
```
BudnikOleksii, line 7: Don't use `any`

src/groupByKey.ts
```diff
@@ -2,6 +2,21 @@ type GroupsMap<T> = {
...
+
+  for (const item of items) {
+    const value = item[key];
+
+    if (!uniqueValues.includes(value)) {
+      uniqueValues.push(value);
+    }
+  }
+
+  uniqueValues.forEach((value) => {
```
BudnikOleksii, line 17: It could be simplified a lot, try to do it without the `uniqueValues` array at all. And actually you can just generate an object with `reduce` method

BudnikOleksii (changes requested):
Good work 👍 
Let's try to simplify your solution, and don't use `any` as it's the same as work without typescript at all

src/groupByKey.ts
```diff
@@ -2,6 +2,16 @@ type GroupsMap<T> = {
...
+  return items.reduce((acc: GroupsMap<T>, item: T) => {
+    const value = String(item[key]);
+
+    if (!acc[value]) {
+      acc[value] = [];
+    }
+
+    acc[value].push(item);
+
+    return { ...acc };
```
BudnikOleksii, line 15: You can just return `acc`

BudnikOleksii (approved):
Looks good 👍

PR #1552, opened 2024-08-20
solutionV1

kgGugala:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+  const result: GroupsMap<T> = {};
+
+  for (const item of items) {
+    const keyValue = item[key];
+
+    if (!((keyValue as string) in result)) {
+      result[keyValue as string] = [];
+    }
+    result[keyValue as string].push(item);
+  }
```
natalia-klonowska, line 15: type assertions is unsafe so ensure that keyValue is string (e.g String()) and delete all "as string"

kgGugala:
I think now everything should be alright if I understood correctly :)

natalia-klonowska (approved):

PR #1561, opened 2024-09-09
task solution

YuliiaKosenchuk:


etojeDenys (approved):
you can simplify it using the reduce method, but looks good

PR #1581, opened 2024-09-29
Solution

v1RnT:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: Array<T>, key: keyof T): GroupsMap<T> {
+  const resOfArr: GroupsMap<T> = {};
```
VolodymyrKirichenko: `resOfArr`? Change this variable name, cuz it is not entirely clear what is contained in the middle.
use `groupedItems` for example.

src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
...
+
+  for (const item of items) {
+    const keyOfT = String(item[key]);
```
VolodymyrKirichenko: `keyOfT`???
key of what? `key of tank`?`key of titanic`?
change it

VolodymyrKirichenko (changes requested):
Good code, but naming problems) 🥹 
Fix it.

PR #1667, opened 2025-01-21
Implement a function groupByKey

yzhyhaliuk:


src/groupByKey.ts
```diff
@@ -2,6 +2,18 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const result: GroupsMap<T> = {};
+
+  for (const item of items) {
```
volodymyr-soltys97, line 8: Consider will be better if you use `forEach` method here

volodymyr-soltys97 (approved):
Good job! 👍

PR #1876, opened 2025-11-04
add task solution 4.11.25

Alya-ya:


src/groupByKey.ts
```diff
@@ -2,6 +2,17 @@ type GroupsMap<T> = {
   [key: string]: T[];
 };
 
-export function groupByKey(items, key) {
-  // write code here;
+export function groupByKey<T>(items: T[], key: keyof T): GroupsMap<T> {
+  const newItem: GroupsMap<T> = {};
```
brespect: 1. Better to rename this like `group `or `result`, 
2. Also you may use here Js Array method `.reduce` to solve this more efficiently

brespect (changes requested):
Nice work, let`s do some changes!

vitalii-bondarenko-ua (approved):
Looks good.

